PR #1, opened 2024-03-07
deploy workflows

skyqrose:
So this is ready for whenever infrastructure exists.

I don't have a way to test it yet.

Major differences from Glides deploy scripts:
- I've removed any reference to sentry (and a couple other things), since it doesn't exist yet.
- version-name is required in all envs. `version_id` is now a descriptive name that we can use for many purposes, including docker tags and sentry releases. It'll look like `2024-03-07:abc123` or `pr:1:abc123` or `main:abc123`, including the git hash.

### DIFF
--- a/.github/workflows/deploy-prod.yml
+++ b/.github/workflows/deploy-prod.yml
@@ -1,0 +1,20 @@
+name: Deploy to Prod
+
+on:
+  release:
+    types:
+      - released
+
+concurrency:
+  group: prod
+  cancel-in-progress: true
+
+jobs:
+  deploy:
+    name: Deploy to Prod
+    uses: ./.github/workflows/deploy.yml
+    with:
+      env: prod
+      # the release name
+      version-name: ${{ github.ref_name }}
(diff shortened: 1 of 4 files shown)


skyqrose:
I just reread everything again now that I understand more about the deploy process. It all looks good now.

Note that the new application checklist recommends using `mbta/workflows/.github/workflows/deploy-ecs.yml@v2`, but Glides had it inlined for some reason (so we could set custom tags? so we could call docker in a different way?). I've kept Glides' approach with more custom code.

skyqrose:
Ready for review. I won't merge it until https://github.com/mbta/devops/pull/1784 and https://github.com/mbta/devops/pull/1785 are merged.

I don't have any plan to test this besides merge it and then see if it works.
If it doesn't work, any future iterations might be able to be tested on a branch instead of requiring a merge.

lemald (approved):
Looks good to me as a base to start experimenting on. 👍

PR #5, opened 2024-05-13
migrate on startup

skyqrose:


mathcolo:
Are these serial?

mathcolo (approved):

skyqrose:
It doesn't specifically say in [the docs](https://hexdocs.pm/ecto_sql/Ecto.Migrator.html#module-example-running-migrations-on-application-startup), but looking at the [source](https://github.com/elixir-ecto/ecto_sql/blob/b4329a1fe6f2888b5b99b3b5b3316f246a838c3a/lib/ecto/migrator.ex#L519), it runs synchronously during `init`, so yes, this will finish before the rest of the application starts.

PR #6, opened 2024-05-20
frontend build process

skyqrose:
Asana Task: A chunk of [Install front-end dev tooling](https://app.asana.com/0/1200689710863995/1207302888667751/f)

- Move `js` and `css` from `assets/` to the root.
  - This goes against the default phoenix organization, but that organization is optimized for mostly phoenix with a little js. We're gonna have a lot of js, and it'd be nice to not need to add `assets/` to every single path I ever write.
  - There's no frontend test directory yet, but I'm planning to put it in `js/test/` and also to put some unit tests side-by-side their source files.
- Installs prettier
  - And runs it. The edits in `README` and `js/phoenix.js` are just formatting.
- Moves `tailwind` and `esbuild` from phoenix to npm.
  - The main motivation is that the prettier-plugin-tailwindcss plugin needs tailwind to be part of npm.
  - But it'll also just be nicer this way, since npm is better set up for managing js/css stuff in general.
  - The main reason to have it in elixir is to avoid including npm in the starter project. But we need npm anyway.
- Adds a frontend build step to `Dockerfile`.

Does not include:
- Updates to `ci.yml` to start running frontend tests. Right now the only test is prettier. I'll add that in a future PR when there's eslint and/or jest.
- eslint / jest / typescript. Those will be in a future PR.
- combining/documenting npm scripts and mix aliases to make them easier to use, like `mix setup` (which is no longer complete) or Glides' `bin/setup`.

- Tests:
  - `( )` Has tests
  - `( )` Doesn't need tests
  - `(x)` Tests deferred (with justification)
- Product/Design sign off:
  - `(x)` No review needed


tailwind.config.ts
```diff
@@ -0,0 +1,103 @@
...
+const fs = require("fs");
+const path = require("path");
+
+module.exports = {
+  content: [
+    "./js/**/*.js",
+    "./js/**/*.ts",
+    "./js/**/*.tsx",
+    "./lib/orbit_web.ex",
+    "./lib/orbit_web/**/*.*ex",
```
skyqrose, line 14: git didn't notice that this file was moved and edited, rather than deleted and recreated.

The only change in this file aside from formatting was this list of places to look for tailwind classes.

Dockerfile
```diff
@@ -15,17 +13,43 @@ RUN mix local.rebar --force
...
+WORKDIR /app
+RUN npm ci
+
+# build frontend
+COPY --from=elixir-deps /app/deps /app/deps
+COPY ./tailwind.config.ts /app/tailwind.config.ts
+COPY ./js /app/js
+COPY ./css /app/css
+COPY ./lib/orbit_web /app/lib/orbit_web
+WORKDIR /app
```
lemald, line 33: **question:** is this redundant with the `WORKDIR` on line 24? Or is there intentional nesting of `/app` directories or something?

lemald (approved):
I think it would be good to resolve that question on the `Dockerfile` before merging, but otherwise LGTM!

Dockerfile
skyqrose: I just noticed that this `COPY WORKDIR RUN COPY WORKDIR RUN` pattern is the same as we do in Glides. Which doesn't mean that it's right, just that it's not completely broken.
https://github.com/mbta/glides/blob/4e7fb2e70d52251fd93b25df1e3a0d9428fd3243/Dockerfile#L45

skyqrose: Discussion: https://mbta.slack.com/archives/GP5QJM98A/p1716495046574599
The COPY isn't broken because the first arg is on the build host, and the 2nd arg is in the container.
The extra WORKDIR is redundant, like doing `cd` twice. But I don't think it's a problem, it's kind of nice to have the working directory listed before each RUN, it makes each `COPY WORKDIR RUN` block a little more standalone.

PR #11, opened 2024-05-31
Install React 

mathcolo:
Asana Task: [Install React and kick it off on application load](https://app.asana.com/0/1206105669438487/1207259534605998/f)

This patch:
- Adds node to `.tool-versions`
- Installs React
- Sets up a trivial Home/Help page set (w/ a test)
- ⚠️ Does _not_ install Storybook. Next PR!

Checklist

- Tests:
  - `(x)` Has tests
  - `( )` Doesn't need tests
- Product/Design sign off:
  - `( )` No review needed


js/app.tsx
```diff
@@ -0,0 +1,6 @@
+import "./phoenix.js";
+import { App } from "./components/app.js";
+import { createRoot } from "react-dom/client";
+
+const root = createRoot(document.getElementById("app") as Element);
```
lemald, line 5: **suggestion:** I slightly prefer writing this as:

```suggestion
const root = createRoot(document.getElementById("app")!);
```

The `!` specifically asserts that the value is non-null, so it makes it a bit clearer that it's narrowing down `Element | null` as opposed to some other `Element | Foo` type that you might also use an `as` assertion on.

mathcolo: Ah, gotcha! Thanks

js/components/home.tsx
```diff
@@ -0,0 +1,12 @@
+import { ReactElement } from "react";
+
+export const Home = (): ReactElement => {
+  return (
+    <div className="text-3xl">
+      🪐
+      <span className="text-[#ED8B00]">O</span>
+      <span className="text-[#DA291C]">r</span>
+      <span className="text-[#003DA5]">b</span>it
```
lemald, line 9: **praise:** beautiful.

lemald:
@mathcolo one general request, I think it would be best to follow the pattern of other apps like Glides and have a `src` directory rather than a `js` directory, just since we're mainly using TypeScript.

js/app.tsx
lemald: Oh, hah, by default the ESLint rules forbid non-null assertions but not `as` assertions, despite the latter being IMO more dangerous. I think adding an ESLint override would be appropriate here.

js/components/home.tsx
mathcolo: This was all @skyqrose, I just brought it over to the react side. I agree, WD Sky!

mathcolo:
@lemald My guess is that came from `phx.new`. Maybe we can move it around in a subsequent PR?

lemald:
> @lemald My guess is that came from `phx.new`. Maybe we can move it around in a subsequent PR?

It probably did, and point taken that we should just move it as part of another PR - as far as React itself goes, I think this looks good.

lemald (approved):

PR #14, opened 2024-06-03
setup and test scripts

skyqrose:
Asana Task: Unspoken part of [Install frontend dev tooling](https://app.asana.com/0/1206604030470713/1207302888667751/f), I guess, now that there's frontend setup and test chores to do.

- Tests:
  - `( )` Has tests
  - `(x)` Doesn't need tests
- Product/Design sign off:
  - `(x)` No review needed


bin/setup
```diff
@@ -0,0 +1,39 @@
+#!/usr/bin/env sh
+
+# Exit if any subcommand fails
+set -e
```
mathcolo, line 4: Can we `set -ex` and remove the echos below?

skyqrose: Whaaat I never knew about `set -x`. That's cool.

Having tried it, it is slightly different than the echos:
- It shows all commands, even the `command -v mise` which is useful if you're debugging exactly what's happening, but is too verbose for navigating all the text scrolling down your terminal.
- It can't do the little bits of editorializing I put in the echos (pointing out that `mix ecto.setup` is in `(dev)`, and combining `local.hex and local.rebar`)

What do you think about adding it to `bin/test`, but keeping the echos in `bin/setup`?

mathcolo: Points taken! That works for me 👍

skyqrose:
Whoops, I asked for a rereview but forgot to actually do the change to `bin/test`. Anyway, it's actually ready for a rereview now.

mathcolo (approved):

PR #20, opened 2024-06-10
Misc Storybook improvements

lemald:
Asana Task: part of prep work for [Badge tap UI](https://app.asana.com/0/1200273269966439/1207259534606012/f)

This just adds some niceties to Storybook now that we'll be using it to actually build things. In particular, I followed [these directions](https://storybook.js.org/recipes/tailwindcss) to add the ability to toggle between light and dark mode (though I'm not sure if we're actually going to have a dark mode for the pilot) and then I based the setup of the color palette on [this answer](https://github.com/storybookjs/storybook/discussions/25483#discussioncomment-8029096).

Screenshot 2024-06-10 at 14 25 13

Checklist

- Tests:
  - `( )` Has tests
  - `(X)` Doesn't need tests
- Product/Design sign off:
  - `(X)` No review needed

### DIFF
--- a/tailwind.config.ts
+++ b/tailwind.config.ts
@@ -17,12 +17,15 @@ export default {
   ],
   darkMode: ["class", '[data-mode="dark"]'],
   theme: {
-    extend: {
-      colors: {
-        brand: "#FD4F00",
-        "mbta-red": "#DA291C",
-        "mbta-orange": "#ED8B00",
-        "mbta-blue": "#003DA5",
+    colors: {
+      "mbta-red": "#DA291C",
+      "mbta-orange": "#ED8B00",
+      "mbta-blue": "#003DA5",
+      black: "#000000",
+      gray: {
+        100: "#EAEAEA",
+        200: "#D9D9D9",
+        300: "#929292",
       },
     },
   },
--- a/tailwind.config.ts
+++ b/tailwind.config.ts
@@ -3,26 +3,31 @@
 
 import fs from "fs";
 import path from "path";
+import formsPlugin from "@tailwindcss/forms";
 import plugin from "tailwindcss/plugin";
 
-module.exports = {
+export default {
   content: [
     "./js/**/*.js",
     "./js/**/*.ts",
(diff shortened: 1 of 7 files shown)


tailwind.config.ts
```diff
@@ -3,25 +3,34 @@
...
-        brand: "#FD4F00",
+    colors: {
+      "mbta-red": "#DA291C",
+      "mbta-orange": "#ED8B00",
+      "mbta-blue": "#003DA5",
+      black: "#000000",
+      gray: {
+        100: "#EAEAEA",
+        200: "#D9D9D9",
+        300: "#929292",
```
skyqrose, line 28: Where do these come from?
Tailwindcss has builtin grays, but they don't line up with these: https://tailwindcss.com/docs/customizing-colors

tailwind.config.ts
```diff
@@ -3,25 +3,34 @@
...
+export default {
   content: [
     "./js/**/*.js",
     "./js/**/*.ts",
     "./js/**/*.tsx",
     "./lib/orbit_web.ex",
     "./lib/orbit_web/**/*.*ex",
+    "./storybook/**/*.html",
   ],
+  darkMode: ["class", '[data-mode="dark"]'],
```
skyqrose, line 18: This is very different from the [custom dark mode plugin](https://github.com/mbta/glides/blob/c368cc69414920254576a55a83deff5273f6473f/assets/tailwind.config.ts#L51 
) that Glides uses, but I guess we'll reconcile those differences if/when we ever add dark mode to the app.

skyqrose (approved):

tailwind.config.ts
lemald: From the Figma designs linked in the Asana ticket.

tailwind.config.ts
lemald: Yeah, I just went with this because it's what the write-up I was looking at used and it seemed pretty standard. I figure we can / will revisit in the future depending on what we decide for the actual implementation.

PR #21, opened 2024-06-10
feat: Keycloak integration

mathcolo:
Asana Task: [Setup Keycloak integration](https://app.asana.com/0/1206105669438487/1207259534606005/f)

Sets up Keycloak auth in Orbit. ~Hold off on merge until infra/terraform changes are in.~

Checklist

- Tests:
  - `(x)` Has tests
  - `( )` Doesn't need tests
- Product/Design sign off:
  - `( )` No review needed


config/prod.exs
```diff
@@ -23,3 +23,16 @@ config :ehmon, :report_mf, {:ehmon, :info_report}
 # which by default uses df flags that aren't available on alpine's busybox.
 # this tells disksup to use different df flags
 config :os_mon, disksup_posix_only: true
+
+# Auth
+config :ueberauth, Ueberauth,
```
skyqrose, line 28: Does `config/prod.exs` need `config :orbit, OrbitWeb.Auth.Guardian` or `config :ueberauth_oidcc` like `dev.exs` and `test.exs` have?

lib/orbit_web/router.ex
```diff
@@ -14,13 +14,27 @@ defmodule OrbitWeb.Router do
     plug :accepts, ["json"]
   end
 
+  pipeline :authenticated do
```
skyqrose, line 17: At the bottom of this file, there's a route for the live dashboard that should also be protected by auth.

test/orbit_web/controllers/auth_controller_test.exs
```diff
@@ -0,0 +1,53 @@
...
+  use OrbitWeb.ConnCase
+  alias OrbitWeb.Auth.Auth
+
+  defp login_with_groups(conn, _groups) do
+    conn
+    |> get(~p"/auth/keycloak/callback?#{%{"email" => "user@example.com"}}")
+  end
+
+  describe "login_page" do
+    test "redirects to keycloak login (i.e. there isn't one)", %{conn: conn} do
```
skyqrose, line 11: huh? I don't understand the parenthetical here. I was expecting something like "/ redirects to keycloak login if you're not logged in".

test/orbit_web/controllers/auth_controller_test.exs
```diff
@@ -0,0 +1,53 @@
+defmodule OrbitWeb.AuthControllerTest do
```
skyqrose, line 1: A case that's not covered here: If you're already logged in, and you try to visit `/login`, it redirects you to `/`.

config/prod.exs
skyqrose: Oh, I think these need to be in `runtime.exs`, which is missing from this PR.

lib/orbit/authentication/user.ex
```diff
@@ -0,0 +1,23 @@
+defmodule Orbit.Authentication.User do
+  use Ecto.Schema
+  import Ecto.Changeset
+
+  @type t :: %__MODULE__{}
```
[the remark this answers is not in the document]
mathcolo: Yep! 0fca10a

lib/orbit_web/auth/auth.ex
```diff
@@ -0,0 +1,57 @@
...
+        Repo.get_by(User, email: email)
+      else
+        nil
+      end
+    end
+  end
+
+  @spec logout(Conn.t()) :: Conn.t()
+  def logout(conn) do
+    # TODO delete refresh token
```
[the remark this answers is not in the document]
mathcolo: 🤦 db788fd

lib/orbit_web/controllers/auth_controller.ex
```diff
@@ -0,0 +1,127 @@
...
+  end
+
+  def callback(conn, _params) do
+    conn
+    |> put_status(:bad_request)
+    |> text("Bad Request (unrecognized provider)")
+  end
+
+  def redirect_needs_login(conn) do
+    if get_format(conn) == "json" do
```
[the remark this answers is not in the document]
mathcolo: 👍 690403a

lib/orbit_web/router.ex
mathcolo: 👍 2bec206

config/prod.exs
mathcolo: 👍 fe54917

test/orbit_web/controllers/auth_controller_test.exs
mathcolo: This was poorly worded. I was insinuating that there is no login _page_ at all in orbit. I switched it to describe `/login` and changed the test description. 17d5e4d

test/orbit_web/controllers/auth_controller_test.exs
mathcolo: This is annoying to implement, as data about being logged in is inserted by the `RequireLogin` plug, but we don't put `/login` behind that plug (for obvious reasons). This doesn't work in Glides either.

skyqrose: 🤯 I didn't realize that was broken on Glides.

I guess this line in `auth_controller.ex` doesn't do anything then.
https://github.com/mbta/orbit/pull/21/files#diff-aa3dda579650ee9ac3c145b3e3254d8dddac9fb940e94016013ea10123fc414aR8

```ex
  def login_page(conn, _params) do
    if Auth.logged_in_user(conn) do
      conn
      |> redirect_login_successful()
```

(I'm not asking you to delete it though.)

mix.exs
```diff
@@ -49,6 +49,8 @@ defmodule Orbit.MixProject do
       {:floki, ">= 0.30.0", only: :test},
       # used by ex_aws
       {:hackney, "== 1.20.1"},
+      {:guardian, "== 2.3.2"},
+      {:guardian_phoenix, "== 2.0.1"},
```
skyqrose, line 53: These aren't alphabetized (should be before `hackney`)

mix.lock
```diff
@@ -1,49 +1,56 @@
 %{
-  "bandit": {:hex, :bandit, "1.2.2", "569fe5d0efb107c9af37a1e37e25ce2ceec293101a2d4bc512876fc3207192b5", [:mix], [{:hpax, "~> 0.1.1", [hex: :hpax, repo: "hexpm", optional: false]}, {:plug, "~> 1.14", [hex: :plug, repo: "hexpm", optional: false]}, {:telemetry, "~> 0.4 or ~> 1.0", [hex: :telemetry, repo: "hexpm", optional: false]}, {:thousand_island, "~> 1.0", [hex: :thousand_island, repo: "hexpm", optional: false]}, {:websock, "~> 0.5", [hex: :websock, repo: "hexpm", optional: false]}], "hexpm", "2f89adb7281c78d4e75733e0a9e1b24f46f84d2993963d6fa57d0eafadec5f03"},
+  "bandit": {:hex, :bandit, "1.5.3", "c7ee44871da696371a5674dd2c2062e974a18cd787732efcf50cc70b98c78fdc", [:mix], [{:hpax, "~> 0.1.1", [hex: :hpax, repo: "hexpm", optional: false]}, {:plug, "~> 1.14", [hex: :plug, repo: "hexpm", optional: false]}, {:telemetry, "~> 0.4 or ~> 1.0", [hex: :telemetry, repo: "hexpm", optional: false]}, {:thousand_island, "~> 1.0", [hex: :thousand_island, repo: "hexpm", optional: false]}, {:websock, "~> 0.5", [hex: :websock, repo: "hexpm", optional: false]}], "hexpm", "3812ed5e48c1f1e3109edb5c463c6f8aaf25ecfac2826606be3e5237550116ef"},
```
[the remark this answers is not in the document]
skyqrose: Made the separate PR: https://github.com/mbta/orbit/pull/24

mix.exs
skyqrose: fixed in [faa8eea](https://github.com/mbta/orbit/pull/21/commits/faa8eea2414044416f2e2a3086988bc29af454b4)

mix.lock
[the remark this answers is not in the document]
skyqrose: Merged that in and removed the extraneous deps updates in [faa8eea](https://github.com/mbta/orbit/pull/21/commits/faa8eea2414044416f2e2a3086988bc29af454b4)

lib/orbit_web/auth/guardian.ex
```diff
@@ -0,0 +1,44 @@
+defmodule OrbitWeb.Auth.Guardian do
+  use Guardian, otp_app: :orbit
+
+  # Guardian expects `resource` to be a full user object.
+  # We don't really have that, so use email address for both id and resource
```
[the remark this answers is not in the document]
lemald: As I understand it from the second of the two code snippets in [this section of the Guardian docs](https://hexdocs.pm/guardian/introduction-implementation.html#basic-setup), the "resource" can basically be whatever struct or other piece of data represents a "user" in your system, so it's really up to us to define what that should be. It's still a pretty simple example, but it might be worth comparing to [how Skate does it](https://github.com/mbta/skate/blob/bacc5520a66f441a3c60772cfbae0611379148a1/lib/skate_web/auth_manager.ex#L30).

lib/orbit_web/controllers/auth_controller.ex
```diff
@@ -0,0 +1,127 @@
...
+      {:ok, logout_url} -> logout_url
+      _ -> nil
+    end
+  end
+
+  def callback(%{assigns: %{ueberauth_auth: %{provider: :keycloak} = auth}} = conn, _params) do
+    username = String.replace(auth.info.email, "MBTA.com", "mbta.com")
+    credentials = auth.credentials
+
+    # Ignore auth provider's TTL, set ours to 30 days so users don't have to log back in
```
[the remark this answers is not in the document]
lemald: Plus one that in general we should try to adopt those guidelines but it doesn't have to be right this moment.

test/orbit_web/controllers/auth_controller_test.exs
lemald: Perhaps we could instead pipe the login path through the [`Guardian.Plug.LoadResource`](https://hexdocs.pm/guardian/Guardian.Plug.LoadResource.html#content) plus with `allow_blank: true`? I think that would make the user information available if present, and otherwise not cause the request to fail.

skyqrose:
Infra PRs:
https://github.com/mbta/terraform_modules/pull/18
https://github.com/mbta/devops/pull/1901

Those will have to be finished before we can actually check that logging in with Keycloak in this PR works.

lib/orbit_web/auth/guardian.ex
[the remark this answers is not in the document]
mathcolo: @lemald What concern led to the introduction of an additional/separate resource prefix for keycloak? It looks like the Skate team was concerned about what would happen around the transition, but I don't us having issues there (a Guardian expiry would trigger a Keycloak/EntraID new session, yes?).

lib/orbit_web/controllers/auth_controller.ex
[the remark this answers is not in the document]
mathcolo: For posterity: https://app.asana.com/0/616151179860796/1207576013276793/f

lib/orbit_web/auth/guardian.ex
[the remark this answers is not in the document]
lemald: You're correct that it would force users to log back in with the new SSO provider. I think it was just a general precautionary measure, see mbta/skate#2400 and the description I wrote there.

The one concrete case I would worry about would be the logout button - what happens to someone who logged in with Cognito when they click logout?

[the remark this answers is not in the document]
mathcolo: In Glides or Orbit? No Cognito here; in Glides, I assume we'll leave the [Cognito logout case](https://github.com/mbta/glides/blob/main/lib/glides_web/controllers/auth_controller.ex#L123) present for a little while after keycloak is defaulted.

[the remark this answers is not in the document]
lemald: Yes, sorry, because you were talking about the transition I thought you were referring to Glides. It isn't really relevant for Orbit at all.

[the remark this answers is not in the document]
mathcolo: Ah, np, yeah I was referring to Skate's transition. Sounds good 👍

test/orbit_web/controllers/auth_controller_test.exs
mathcolo: Hmm. We're already using `allow_blank` in `OrbitWeb.Auth.Guardian.Pipeline`, and I'm not sure what using LoadResource's `allow_blank` _without_ `VerifySession` does from a security standpoint. Perhaps I'll just pull this functionality from `login_page/2` and log a task?

Because /login on Orbit is an immediate redirect to `/auth/keycloak`, it has the benefit of looking like a redirect to `/` even if slow.

config/prod.exs
```diff
@@ -4,7 +4,8 @@ import Config
 
 config :orbit,
   # loaded at compile time
-  release: System.get_env("RELEASE")
+  release: System.get_env("RELEASE"),
+  force_https?: true
```
lemald, line 8: **suggestion:** I'd have a slight preference for setting `force_https?: true` in `config.exs` and only setting it to `false` explicitly in `dev.exs` just to be a bit safer.

lib/orbit_web/auth/auth.ex
```diff
@@ -0,0 +1,56 @@
...
+  alias Orbit.Authentication.User
+  alias Orbit.Repo
+  alias Plug.Conn
+
+  @spec login(Conn.t(), String.t(), integer(), String.t() | nil, [String.t()], String.t() | nil) ::
+          Conn.t()
+  def login(conn, username, ttl_seconds, _refresh_token, _groups, logout_url) do
+    email = String.downcase(username)
+
+    Repo.transaction(fn ->
```
lemald, line 11: **question:** what's the purpose of the transaction here? If it's to avoid having a record inserted with the same username between the `Repo.get_by/2` and the `Repo.insert!/1`, I don't believe the standard isolation level of Postgres transactions will accomplish that.

lib/orbit_web/controllers/auth_controller.ex
```diff
@@ -0,0 +1,127 @@
...
+  use OrbitWeb, :controller
+  plug(Ueberauth)
+
+  alias OrbitWeb.Auth.Auth
+  require Logger
+
+  def login_page(conn, _params) do
+    if Auth.logged_in_user(conn) do
+      conn
+      |> redirect_login_successful()
```
lemald, line 11: **nitpick:** we usually don't pipeline function calls if it results in just a single pipeline step.

lib/orbit_web/controllers/auth_controller.ex
```diff
@@ -0,0 +1,127 @@
...
+    else
+      conn |> redirect(to: ~p"/auth/keycloak")
```
lemald, line 13: **nitpick:** see above comment about pipeline operator.

lib/orbit_web/controllers/auth_controller.ex
```diff
@@ -0,0 +1,127 @@
...
+    conn
+    |> put_status(:bad_request)
+    |> text("Bad Request (unrecognized provider)")
+  end
+
+  defp get_logout_url(conn, auth) do
+    case UeberauthOidcc.initiate_logout_url(auth, %{
+           # We have to generate exactly what Keycloak/EntraID expect for logout redirect
+           post_logout_redirect_uri:
+             URI.parse(url(conn, ~p"/login"))
```
lemald, line 29: **nitpick:** if you're pipelining the rest of this, you should also pull this out (formatting as a suggestion but you probably shouldn't commit it directly because that would break the formatting).

```suggestion
             conn |> url(~p"/login") |> URI.parse()
```

lib/orbit_web/controllers/auth_controller.ex
```diff
@@ -0,0 +1,127 @@
...
+
+  def callback(%{assigns: %{ueberauth_auth: %{provider: :keycloak} = auth}} = conn, _params) do
+    username = String.replace(auth.info.email, "MBTA.com", "mbta.com")
+    credentials = auth.credentials
+
+    # Ignore auth provider's TTL, set ours to 30 days so users don't have to log back in
+    # expiration = credentials.expires_at
+    ttl_seconds = 3600 * 24 * 30
+
+    refresh_token = Map.get(credentials, :refresh_token, nil)
```
lemald, line 50: **suggestion:** since it appears that we aren't doing anything with refresh tokens yet (the argument to `Auth.login/5` is ignored), it would probably be better to just remove this value and the corresponding function argument in order to avoid confusion.

test/orbit_web/controllers/auth_controller_test.exs
lemald: My understanding is that `LoadResource` with `allow_blank: true` and none of the other plugs like `VerifySession` or `VerifyHeader` would load the resource, if present, and otherwise do nothing. That would give us access to the user resource without necessarily implying that the user is correctly logged in. That said, as I think about this more I'm not sure how we would avoid potential redirect loops if we found a logged in user but then the page we redirected them to redirected right back to the login flow because the login isn't actually valid. All of that said, I think I'd be fine with removing that code from the PR for now and creating a followup ticket; it's not super-critical functionality.

lemald:
@mathcolo okay, I'm done with a full pass of review and left some comments for you to check out!

lib/orbit_web/auth/auth.ex
mathcolo: Yeah, I think you're right. This is also in Glides, so I'll remove it here and then there in a separate PR 👍

lib/orbit_web/auth/auth.ex
```diff
@@ -0,0 +1,51 @@
...
+  def login(conn, username, ttl_seconds, _groups, logout_url) do
+    email = String.downcase(username)
+
+    case Repo.get_by(User, email: email) do
+      nil ->
+        Repo.insert!(%User{email: email})
+
+      u ->
+        u
+    end
```
lemald, line 17: **suggestion:** would this work as a simpler version of the same logic?
```suggestion
    Repo.get_by(User, email: email) || Repo.insert!(%User{email: email})
```

lib/orbit_web/controllers/auth_controller.ex
```diff
@@ -0,0 +1,124 @@
...
+             |> URI.to_string()
+         }) do
+      {:ok, logout_url} -> logout_url
+      _ -> nil
+    end
+  end
+
+  def callback(%{assigns: %{ueberauth_auth: %{provider: :keycloak} = auth}} = conn, _params) do
+    username = String.replace(auth.info.email, "MBTA.com", "mbta.com")
+    # credentials = auth.credentials
```
lemald, line 45: **nitpick:** remaining commented-out code.

lib/orbit_web/auth/auth.ex
lemald: Thanks for making the change here and in Glides! It does still leave the question of how to best handle the race condition, but it's not likely to happen very often at all (it would require someone logging into the app for the first time roughly simultaneously from two devices or browsers), and if it ever does happen someone would just have to refresh and then the next request should work correctly, so I'm not too worried about it.

lib/orbit_web/controllers/auth_controller.ex
```diff
@@ -0,0 +1,124 @@
...
+
+    conn
+    |> Auth.logout()
+    |> redirect_needs_login()
+  end
+
+  def callback(conn, _params) do
+    conn
+    |> put_status(:bad_request)
+    |> text("Bad Request (unrecognized provider)")
```
skyqrose, line 89: "unrecognized provider" isn't really accurate at this point.
```suggestion
    |> text("Bad Request")
```

skyqrose (approved):

lib/orbit_web/controllers/auth_controller.ex
mathcolo: I was hoping to leave this in whenever we get to keycloak groups/refresh tokens. But I can take it out!

lib/orbit_web/controllers/auth_controller.ex
mathcolo: Ugh. Seriously...

PR #28, opened 2024-06-20
feat: Add Oban

mathcolo:
Asana Task: Supporting [Import basic operator data into Orbit backend](https://app.asana.com/0/1206105669438487/1207392635878919/f)

Checklist

- Tests:
  - `( )` Has tests
  - `(x)` Doesn't need tests
- Product/Design sign off:
  - `(x)` No review needed


mix.exs
```diff
@@ -67,7 +68,8 @@ defmodule Orbit.MixProject do
       {:postgrex, ">= 0.0.0"},
       {:sobelow, "== 0.13.0", only: [:dev, :test], runtime: false},
       {:telemetry_metrics, "== 1.0.0"},
-      {:telemetry_poller, "== 1.1.0"}
+      {:telemetry_poller, "== 1.1.0"},
+      {:timex, "== 3.7.11"}
```
skyqrose, line 72: What's timex used for?

skyqrose (commented):
I see glides has this line in `config/test.exs`:
```
config :glides, Oban, testing: :inline
```
Seems like Orbit probably should, too.

https://hexdocs.pm/oban/testing.html
> Both testing modes prevent Oban from running any database queries in the background. This simultaneously prevents Sandbox errors from plugin queries and prevents queues from executing jobs unexpectedly.

I don't think there's any other test config we'd need until we actually write a test that uses it.

mix.exs
mathcolo: Good question- `timex` depends on `tzdata`, which is needed for us to be able to give the named zone `America/New_York` to Oban's "cron" plugin. This matches what's in glides, but it would be more correct for us to depend directly on `tzdata`, and only optionally add `timex` later.

skyqrose (approved):

PR #29, opened 2024-06-21
feat: basic modal component

lemald:
Asana Task: initial work on [Badge tap UI](https://app.asana.com/0/1200273269966439/1207259534606012/f)

This is just a very simple modal component, and it probably doesn't work well on mobile yet. I didn't opt to try to put this in Storybook since it's sort of weird to add modals there, but I'm splitting this out into its own component so that I can make a separate component for the contents that will go into Storybook.

Checklist

- Tests:
  - `(X)` Has tests
  - `( )` Doesn't need tests
- Product/Design sign off:
  - `(X)` No review needed

### DIFF
--- a/js/components/modal.tsx
+++ b/js/components/modal.tsx
@@ -1,0 +1,36 @@
+import { Modal as RestartModal } from "@restart/ui";
+import { ReactElement } from "react";
+
+export const Modal = ({
+  children,
+  title,
+  show,
+  onHide,
+}: {
+  children: ReactElement;
+  title: ReactElement;
+  show: boolean;
+  onHide: () => void;
+}): ReactElement => {
+  return (
+    <RestartModal
+      show={show}
+      onHide={onHide}
+      className="fixed z-modal-content flex max-h-[calc(100vh-3.5rem)] flex-col bg-gray-100 border-2 border-black p-1 -translate-x-2/4 -translate-y-2/4 top-2/4 left-2/4"
+      renderBackdrop={(props) => (
+        <div
+          {...props}
+          className="fixed inset-0 z-modal-backdrop w-full h-full bg-black/[0.48]"
+        />
+      )}
+    >
+      <div className='grid grid-cols-[1fr_max-content] grid-rows-[max-content_1fr] [grid-template-areas:"title_close"_"contents_contents"]'>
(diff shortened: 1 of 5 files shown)


tailwind.config.ts
```diff
@@ -28,6 +28,12 @@ export default {
         300: "#929292",
       },
     },
+    zIndex: {
+      layout: "1000",
+      object: "2000",
```
skyqrose, line 33: What are `z-layout` and `z-object` used for?

skyqrose (approved):

tailwind.config.ts
lemald: At the moment they're just placeholders by analogy to some of the named z-index values in Glides.

js/components/modal.tsx
```diff
@@ -0,0 +1,36 @@
+import { Modal as RestartModal } from "@restart/ui";
+import { ReactElement } from "react";
+
+export const Modal = ({
+  children,
+  title,
+  show,
+  onHide,
```
mathcolo, line 8: Might be more clear as `onClose`?

js/components/modal.tsx
```diff
@@ -0,0 +1,36 @@
...
+        <div
+          {...props}
+          className="fixed inset-0 z-modal-backdrop w-full h-full bg-black/[0.48]"
+        />
+      )}
+    >
+      <div className='grid grid-cols-[1fr_max-content] grid-rows-[max-content_1fr] [grid-template-areas:"title_close"_"contents_contents"]'>
+        <div className="[grid-area:title]">{title}</div>
+        <button className="[grid-area:close]" onClick={onHide}>
+          [x]
```
mathcolo, line 30: I'm partial to no square brackets, but 🤷

mathcolo (approved):

js/components/modal.tsx
lemald: I went with `onHide` because that's what Restart UI uses, but I think you're right that `onClose` is clearer. 54f1b2b

js/components/modal.tsx
lemald: Yeah, my preference is that we just get the icons working eventually and then ditch this version entirely.

PR #33, opened 2024-07-02
Operator sign in schema

lemald:
Asana Task: [Database schema for operator sign in history](https://app.asana.com/0/1200273269966439/1207259531016320/f)

This implements the basic schema for an `operator_sign_ins` table and some of the Elixir tooling to go with it, including ExMachina factories and a simple test that shows how we can easily work with associations and foreign key constraints.

Checklist

- Tests:
  - `(X)` Has tests
  - `( )` Doesn't need tests
- Product/Design sign off:
  - `(X)` No review needed


lib/orbit/operator_sign_in.ex
```diff
@@ -0,0 +1,35 @@
...
+  use Ecto.Schema
+  import Ecto.Changeset
+
+  alias Orbit.Employee
+  alias Orbit.RailLine
+  alias Orbit.SignInMethod
+  alias Orbit.User
+
+  @type t :: %__MODULE__{
+          signed_in_employee: Employee.t(),
```
mathcolo, line 11: **suggestion (low)**: The `signed_in` prefixes here are kind of redundant.

lib/orbit/operator_sign_in.ex
```diff
@@ -0,0 +1,35 @@
...
+          signed_in_by_user: User.t(),
+          signed_in_at: DateTime.t(),
+          rail_line: RailLine.t(),
```
mathcolo, line 14: suggestion (medium): Should this be `route`?

lib/orbit/rail_line.ex
```diff
@@ -0,0 +1,20 @@
+defmodule Orbit.RailLine do
+  use Ecto.Type
+
+  @type t :: :red | :orange | :blue
+
+  @impl true
+  def type, do: :string
+
+  @impl true
+  def cast(line) when line in [:red, :orange, :blue], do: {:ok, line}
```
mathcolo, line 10: Is there no need for a cast from string?

lib/orbit/user.ex
```diff
@@ -0,0 +1,14 @@
+defmodule Orbit.User do
```
mathcolo, line 1: **question (high):** Is this branch rebased against `main`? `main` has a `Orbit.Authentication.User`.

mix.exs
```diff
@@ -113,7 +114,7 @@ defmodule Orbit.MixProject do
       test_all: [
         "test",
         "credo",
-        "format --check-formatted",
+        "Format --check-formatted",
```
mathcolo, line 117: **question (low): ** what was this for?

priv/repo/migrations/20240701160107_add_operator_sign_ins.exs
```diff
@@ -0,0 +1,28 @@
+defmodule Orbit.Repo.Migrations.AddOperatorSignIns do
+  use Ecto.Migration
+
+  def change do
+    execute(
+      "CREATE TYPE rail_line AS ENUM ('red', 'orange', 'blue')",
+      "DROP TYPE rail_line"
```
mathcolo, line 7: **praise: ** I didn't know executes could take a rollback statement as the second argument. nice!

mix.exs
lemald: Ugh, sorry, that was an error that I thought I removed in a fixup commit. Should be actually fixed in c9baca6.

lib/orbit/rail_line.ex
lemald: The [`cast/1` callback](https://hexdocs.pm/ecto/Ecto.Type.html#c:cast/1) here is used as part of the Ecto changeset "cast" operation, it basically just validates. Actually turning it into the value that gets sent to the DB is the job of the `dump/1` callback.

lib/orbit/user.ex
lemald: It is based off of the latest `main`, I just didn't realize that module already existed. 🤦 I'll remove my duplicate implementation.

priv/repo/migrations/20240701160107_add_operator_sign_ins.exs
lemald: Yes, it's quite handy! Lets you intersperse `execute` with other migration operations without having to then separate things out into a separate `up` and `down` just because of a single raw SQL statement you need.

lib/orbit/user.ex
lemald: Okay fixed in ab2dfbd (see also f776841 for another tiny improvement).

lib/orbit/operator_sign_in.ex
lemald: Good question. I didn't give it much thought initially but as I weigh it I think "line" is more appropriate than "route," because "route" is already a concept in GTFS and I'm thinking of this as a concept that's agnostic to exactly how routes are modeled in GTFS. For instance, the Green Line is actually 4 GTFS routes but we'd call it only one "line," and I suppose you could imagine a world where the two Red Line branches are actually separate GTFS routes. Of course, when it comes to the former that's not a concern for Orbit, and for the latter I haven't heard of any future intention of modeling the Red Line that way, so maybe it's not very sound logic after all.

lib/orbit/operator_sign_in.ex
lemald: I wanted to make it clear exactly what the relationship between the `operator_sign_in` record and each party is, especially because in the future we might have individuals who both exist as entries in `users` and `employees`, which I believe is already the case for Glides. My feeling is that it may not be obvious to someone coming to this code for the first time that the "user" is the person doing the signing in and the "employee" is the person being signed in.

lib/orbit/rail_line.ex
mathcolo: Gotcha. Since it's effectively an identity function, could it be omitted?

lib/orbit/operator_sign_in.ex
mathcolo: Fair enough! Cannot argue with clarity 😃

mathcolo (commented):
Should be able to approve this once the conflicts are addressed!

lib/orbit/rail_line.ex
lemald: It's a required callback of the [`Ecto.Type`](https://hexdocs.pm/ecto/Ecto.Type.html#c:cast/1) behaviour, so if we don't define it, Elixir will complain. It's an identity when actually given one of the allowed values, but its real purpose is populating a useful error message if appropriate.

mathcolo (approved):
🚀 if it fits, it ships. With an 0700-sized flat rate train.

PR #35, opened 2024-07-05
feat: useNfc hook

lemald:
Asana Task: **[Badge tap action via NFC API](https://app.asana.com/0/1200273269966439/1207259534606015/f)**

This borrows a lot from the [NFC code in Glides](https://github.com/mbta/glides/blob/097cb9f7ffb9f8d4fccf29272a29c321ca8014a5/assets/src/components/operatorSignIn/nfc.tsx), but puts it in a hook to make it a bit more React-y. One other difference is that the bug with being unable to cancel scans that was an issue at the time the Glides NFC functionality was implemented is no longer an issue on current versions of Chrome.

I have done some manual testing by simply rendering the output of the `useNfc` hook onto the page and confirming that it functions as expected. The `useNfc` hook only scans a card once each time it's mounted and then cancels the scan, and unmounting the hook takes care of properly tearing down the scan.

The main issue at the moment is that I still have only partial automated testing for this. It turns out that working with and mocking globally-defined classes in Jest is _really_ difficult. I think especially for something as tricky as this feature automated testing is especially important, and I have managed to test at least a couple of important cases, but we'll want to continue to refine the tests as we go.

Checklist

- Tests:
  - `(X)` Has tests
  - `( )` Doesn't need tests
  - `(X)` Tests deferred (with justification)
- Product/Design sign off:
  - `( )` No review needed

### DIFF
--- a/js/hooks/useNfc.ts
+++ b/js/hooks/useNfc.ts
@@ -54,5 +54,5 @@ const dehexSerial = (hexedSerial: string): string => {
   const bytes = hexedSerial.split(":");
   const hexValue = bytes.join("");
   const decimalValue = parseInt(hexValue, 16);
-  return `${decimalValue}`;
+  return decimalValue.toString();
 };
--- a/js/hooks/useNfc.ts
+++ b/js/hooks/useNfc.ts
@@ -1,51 +1,59 @@
+import { nfcSupported } from "../util/nfc";
 import { useEffect, useState } from "react";
 
 export type NfcResult =
+  | { status: "nfcUnsupported" }
   | { status: "reading" }
   | { status: "success"; data: string }
   | { status: "cancelled" }
   | { status: "error"; error: Error };
 
 export const useNfc = (cancel: AbortController): NfcResult => {
-  const [result, setResult] = useState<NfcResult>({ status: "reading" });
+  const supported = nfcSupported();
+
+  const [result, setResult] = useState<NfcResult>({
+    status: supported ? "reading" : "nfcUnsupported",
+  });
 
   useEffect(() => {
-    const reader = new NDEFReader();
+    if (supported) {
+      const reader = new NDEFReader();
 
-    reader.scan({ signal: cancel.signal }).catch((error: unknown) => {
-      if (error instanceof Error) {
-        setResult({ status: "error", error });
-      }
-    });
-
-    reader.addEventListener(
-      "reading",
-      (event: Event) => {
-        if (event instanceof NDEFReadingEvent) {
-          setResult({
-            status: "success",
-            data: dehexSerial(event.serialNumber),
-          });
-          cancel.abort();
-        }
-      },
-      { once: true },
-    );
-
-    cancel.signal.addEventListener("abort", () => {
-      setResult((result) => {
-        if (result.status === "success") {
-          return result;
-        } else {
-          return { status: "cancelled" };
+      reader.scan({ signal: cancel.signal }).catch((error: unknown) => {
+        if (error instanceof Error) {
+          setResult({ status: "error", error });
         }
       });
-    });
 
-    return () => {
-      cancel.abort();
-    };
-  }, [cancel]);
+      reader.addEventListener(
+        "reading",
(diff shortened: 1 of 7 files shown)


js/hooks/useNfc.ts
```diff
@@ -0,0 +1,58 @@
+import { useEffect, useState } from "react";
+
+export type NfcResult =
+  | { status: "reading" }
+  | { status: "success"; data: string }
+  | { status: "cancelled" }
+  | { status: "error"; error: Error };
+
+export const useNfc = (cancel: AbortController): NfcResult => {
+  const [result, setResult] = useState<NfcResult>({ status: "reading" });
```
mathcolo, line 10: It seems incorrect that on mount you'd start in `reading` state, even if `NDEFReader` is either unavailable or not working properly.

js/hooks/useNfc.ts
```diff
@@ -0,0 +1,58 @@
...
+  }, [cancel]);
+
+  return result;
+};
+
+const dehexSerial = (hexedSerial: string): string => {
+  const bytes = hexedSerial.split(":");
+  const hexValue = bytes.join("");
+  const decimalValue = parseInt(hexValue, 16);
+  return `${decimalValue}`;
```
mathcolo, line 57: Would a `toString` use be better here?

js/hooks/useNfc.ts
lemald: I was working off of the Glides NFC code and included the `nfcSupported` utility function so that we can determine whether to route the user to the NFC-based flow or not, so this hook shouldn't be used from the non-NFC-flow components. That said, I could also modify things such that the `NfcResult` type instead has an `"nfcUnsupported"` status or something and handle that case in the hook itself. What do you think of that?

js/hooks/useNfc.ts
lemald: Yeah, this is one of those things I copied over from Glides but that would probably be a better approach. Changed in 31a179d.

js/hooks/useNfc.ts
lemald: Okay, I actually gave it a shot in faf48e6 - let me know what you think.

mathcolo: Because of what we discussed re: hooks, I think this is an improvement!

js/hooks/useNfc.ts
```diff
@@ -0,0 +1,66 @@
...
+  const [result, setResult] = useState<NfcResult>({
+    status: supported ? "reading" : "nfcUnsupported",
+  });
+
+  useEffect(() => {
+    if (supported) {
+      const reader = new NDEFReader();
+
+      reader.scan({ signal: cancel.signal }).catch((error: unknown) => {
+        if (error instanceof Error) {
```
mathcolo, line 23: This would work fine if it was a String still, right?

js/hooks/useNfc.ts
```diff
@@ -0,0 +1,66 @@
...
+          setResult({ status: "error", error });
+        }
+      });
+
+      reader.addEventListener(
+        "reading",
+        (event: Event) => {
+          if (event instanceof NDEFReadingEvent) {
```
mathcolo, line 31: Should we throw if it's not an `NDEFReadingEvent`?

js/hooks/useNfc.ts
lemald: Yeah, the way Glides does it, the NFC tap functionality is associated with a component, and while you can't conditionally call a hook you _can_ conditionally render a component. But I guess we shouldn't lock ourselves into this approach in Orbit.

js/hooks/useNfc.ts
lemald: Yeah, thinking about it it doesn't really have to be any type in particular. See 18cd5b2.

js/hooks/useNfc.ts
lemald: Good idea, see 18cd5b2.

mathcolo (approved):
Let's do it!

PR #36, opened 2024-07-05
RFID Serial Schema

lemald:
Asana Task: [Badge tap action via NFC API](https://app.asana.com/0/1200273269966439/1207259534606015/f)

I realized this didn't necessarily need to wait for the other ticket that's focused on the import, so I did it as part of this task (and removed the corresponding acceptance criterion from the import ticket).

This makes heavy use of the associations feature in Ecto so that once can get / update employee and modify the `badge_serials` value on the `Employee` struct and have the result reflected correctly in the database. It's similar to (albeit simpler than) the RouteTab functionality in Skate.

One addition thing to note is that I took the opportunity to have the migration update the `on_delete` functionality for the associations on the `operator_sign_ins` table, so that we don't have dangling `operator_sign_in` entries if the corresponding employee or user is deleted (we can always add soft delete functionality if we want to "delete" a user or employee without losing old records).

Checklist

- Tests:
  - `(X)` Has tests
  - `( )` Doesn't need tests
- Product/Design sign off:
  - `(X)` No review needed

### DIFF
--- a/lib/orbit/badge_serial.ex
+++ b/lib/orbit/badge_serial.ex
@@ -1,0 +1,24 @@
+defmodule Orbit.BadgeSerial do
+  use Ecto.Schema
+  import Ecto.Changeset
+
+  alias Orbit.Employee
+
+  @type t :: %__MODULE__{
+          badge_serial: String.t()
+        }
+
+  schema "badge_serials" do
+    belongs_to(:employee, Employee)
+    field(:badge_serial, :string)
+
+    timestamps()
+  end
+
+  def changeset(badge_serial, attrs \\ %{}) do
+    badge_serial
+    |> cast(attrs, [:badge_serial])
+    |> validate_required([:badge_serial])
+    |> foreign_key_constraint(:employee_id)
+  end
+end
--- a/lib/orbit/badge_serial.ex
+++ b/lib/orbit/badge_serial.ex
@@ -5,7 +5,8 @@ defmodule Orbit.BadgeSerial do
   alias Orbit.Employee
 
   @type t :: %__MODULE__{
-          badge_serial: String.t()
+          badge_serial: String.t(),
+          employee: Employee.t() | nil
         }
 
   schema "badge_serials" do
--- a/priv/repo/migrations/20240705175329_add_badge_serials.exs
+++ b/priv/repo/migrations/20240705175329_add_badge_serials.exs
@@ -9,6 +9,8 @@ defmodule Orbit.Repo.Migrations.AddBadgeSerials do
       timestamps()
     end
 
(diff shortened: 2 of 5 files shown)


lib/orbit/badge_serial.ex
```diff
@@ -0,0 +1,24 @@
+defmodule Orbit.BadgeSerial do
+  use Ecto.Schema
+  import Ecto.Changeset
+
+  alias Orbit.Employee
+
+  @type t :: %__MODULE__{
+          badge_serial: String.t()
```
mathcolo, line 8: Should this `@type` contain an in-struct reference to `Employee`, since it's in the schema?

priv/repo/migrations/20240705175329_add_badge_serials.exs
```diff
@@ -0,0 +1,22 @@
+defmodule Orbit.Repo.Migrations.AddBadgeSerials do
+  use Ecto.Migration
+
+  def change do
+    create table("badge_serials") do
+      add :employee_id, references("employees", on_delete: :delete_all)
```
mathcolo, line 6: Realistically we've never going to actually delete from `employees`, right?

lemald: Yeah, if we need something akin to "delete" functionality in the application we can always make a soft delete, but the table entry would stick around for the purposes of having a record of past sign-ins and stuff.

lib/orbit/badge_serial.ex
lemald: I was sort of torn on this one. The association goes both ways, but I think in practice the right way to use these is going to be to reference the `Employee` primarily and look at the `Employee`'s `badge_serials` value, rather than the other way around. Anyhow, all of that said, I don't think it hurts anything to add it as optional, see 7af289d.

priv/repo/migrations/20240705175329_add_badge_serials.exs
```diff
@@ -0,0 +1,22 @@
...
+      add :badge_serial, :string
+
+      timestamps()
+    end
+
+    create index(:badge_serials, [:badge_serial])
+
+    alter table("operator_sign_ins") do
+      modify :signed_in_employee_id, references("employees", on_delete: :delete_all),
```
mathcolo, line 15: **Comment (probably not actionable):** Because we want to keep old sign-ins, it's a _teeny_ bit weird we have a `delete_all` here. Even though I get why.

lemald: Yeah, it's sort of counterintuitive.

mathcolo (approved):

PR #37, opened 2024-07-09
Front-end basics for making API calls

lemald:
Asana Task: arbitrarily started as part of [Badge tap action via NFC API](https://app.asana.com/0/1200273269966439/1207259534606015/f)

@mathcolo sorry if any of this overlaps with your work on the [other part of the operator sign-in flow that will also have to get data from the backend](https://app.asana.com/0/1200273269966439/1207373383816236/f). I didn't go too far with my work so that if there's duplication we can sync up and take care of resolving it without too much effort.

Also one quick note that this is currently based off of the `lem-nfc-basics` branch because of some test utilities. Once #35 is merged I'll rebase this onto `main`.

Checklist

- Tests:
  - `(X)` Has tests
  - `( )` Doesn't need tests
- Product/Design sign off:
  - `(X)` No review needed

### DIFF
--- a/js/api.ts
+++ b/js/api.ts
@@ -1,0 +1,49 @@
+import { fetch } from "./browser";
+import { useEffect, useState } from "react";
+import { z } from "zod";
+
+export type ApiResponse<OkResult, ErrorResult = never> =
+  | { status: "ok"; result: OkResult }
+  | {
+      status: "error";
+      statusCode?: number;
+      result?: ErrorResult;
+    };
+
+export type ApiResult<OkResult, ErrorResult = never> =
+  | { status: "loading" }
+  | ApiResponse<OkResult, ErrorResult>;
+
+export const useApiResult = <RawData, Data>({
+  RawData,
+  url,
+  parser,
+}: {
+  RawData: z.ZodType<RawData>;
+  url: string;
+  parser: (data: RawData) => Data;
+}): ApiResult<Data> => {
+  const [result, setResult] = useState<ApiResult<Data>>({ status: "loading" });
+
+  useEffect(() => {
+    fetch(url)
+      .then((response) => {
+        if (response.status === 200) {
+          return response.json();
+        } else {
+          // TODO: handle various errors
+        }
+      })
+      .then((json) => {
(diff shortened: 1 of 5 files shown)


js/browser.ts
```diff
@@ -0,0 +1,5 @@
+export const fetch: typeof window.fetch = window.fetch;
```
mathcolo, line 1: Can TypeScript not narrow/infer (whatever we call it?) these?

js/api.ts
```diff
@@ -0,0 +1,49 @@
...
+    fetch(url)
+      .then((response) => {
+        if (response.status === 200) {
+          return response.json();
+        } else if (response.status === 401) {
+          reload();
+
+          throw new Error("Unauthenticated");
+        } else {
+          // TODO: handle various errors
```
mathcolo, line 31: In the meantime, could we throw an `Unhandled response code` error or something?

js/browser.ts
lemald: Looks like it can, I forget why I originally included that. ecbdcac

js/api.ts
lemald: Good point. I've made some improvements in 6f8d9b8. Note that throwing inside a promise just causes it to be rejected and the `throw` doesn't bubble up (which, no pun intended, threw me off when I was writing tests).

js/api.ts
```diff
@@ -0,0 +1,49 @@
...
+        const jsonObject = z.object({ data: z.unknown() }).parse(json);
+
+        const data = RawData.parse(jsonObject.data);
+
+        const parsedData = parser(data);
+
+        setResult({ status: "ok", result: parsedData });
+      })
+      .catch((e: unknown) => {
+        setResult({ status: "error", error: e });
```
mathcolo, line 44: Could we `console.error` this just so we see it as well?

mathcolo (approved):
One comment above, but it's non-blocking.

js/api.ts
lemald: I actually think sending an error to Sentry would be more useful, so I added a note to [the ticket](https://app.asana.com/0/616151179860796/1207620537833336/f) for frontend Sentry support. Does that sounds good to you?

PR #32, opened 2024-07-01
feat: Import basic operator data

mathcolo:
Asana Task: [Import basic operator data into Orbit backend](https://app.asana.com/0/1206105669438487/1207392635878919/f)

- Oban worker to import personnel data every morning at 3:50AM.
- Add preferred first name to `employees`.

Checklist

- Tests:
  - `(x)` Has tests
  - `( )` Doesn't need tests
  - `( )` Tests deferred (with justification)
- Product/Design sign off:
  - `(x)` Okayed the plan for the feature (e.g. the design files, or the Asana task)
  - `( )` Reviewed the feature as implemented (e.g. on dev-green, or saw screenshots)
  - `( )` No review needed

PR #40, opened 2024-07-11
Import badge RFID data

lemald:
Asana Task: [Get operator badge information into Orbit backend](https://app.asana.com/0/1200273269966439/1207259531016324/f)

This is heavily based on the work in #32. I think in theory we could refactor things to have a unified import behaviour or something but I don't think we know quite enough about our future use cases to make that worthwhile at this particular moment.

Checklist

- Tests:
  - `(X)` Has tests
  - `( )` Doesn't need tests
- Product/Design sign off:
  - `(X)` No review needed


lib/orbit/import/rfid.ex
```diff
@@ -0,0 +1,75 @@
...
+
+              employee ->
+                {:ok, employee} =
+                  employee
+                  |> Repo.preload([:badge_serials])
+                  |> Employee.changeset(%{
+                    badge_serials:
+                      Enum.map(
+                        badge_serials,
+                        &%{badge_serial: &1[@badge_serial_field]}
```
mathcolo, line 45: This will still work if the serial number is an empty string, right?

test/orbit/import/rfid_test.exs
```diff
@@ -0,0 +1,90 @@
...
+  alias Orbit.BadgeSerial
+  alias Orbit.Employee
+  use Orbit.DataCase
+  alias Orbit.Import.Rfid
+  alias Orbit.Import.RfidWorker
+  import Orbit.Factory
+
+  describe "worker" do
+    test "downloads and parses badge serials" do
+      insert(:employee, %{badge_number: "1234", badge_serials: []})
```
mathcolo, line 11: Could "1234" be a module constant?

lib/orbit/import/rfid.ex
lemald: I think it would just insert a row with empty string, but from looking at the file there aren't any blanks in either of the two columns we're using here.

test/orbit/import/rfid_test.exs
lemald: It maybe could, but not all of the tests are quite the same in terms of what IDs they use (note that there's one test that has two IDs).

lib/orbit/import/rfid.ex
```diff
@@ -0,0 +1,75 @@
...
+        Keyword.fetch!(Application.fetch_env!(:orbit, Orbit.Import.Rfid), :pathname)
+      )
+
+    csv_data
+  end
+
+  @badge_number_field "EMPLOYEE_ID"
+  @badge_serial_field "SERIAL_NUMBER"
+
+  @spec import_rows(Enumerable.t()) :: :ok
```
mathcolo, line 21: Since we know what kind of Enumerable this is, can we specify?

mathcolo (approved):
One additional comment. Otherwise I think we're good to go!

lib/orbit/import/rfid.ex
lemald: The typespec of `CSV.decode!` doesn't get more specific than `Enumerable.t()`, so I think this is the strongest we can get without making additional assumptions about the file format that should be handled at runtime rather than via types.

mathcolo (approved):
Yep, the `:replace_all_except` looks fine!

PR #43, opened 2024-07-12
NFC-based operator lookup

lemald:
Asana Task: [Badge tap action via NFC API](https://app.asana.com/0/1200273269966439/1207259534606015/f)

I'm still working on figuring out a good way to manually test this whole flow, but given that I have what should be fairly complete logic for the happy path put together with automatic tests I figured I'd put a PR up.

Checklist

- Tests:
  - `(X)` Has tests
  - `( )` Doesn't need tests
- Product/Design sign off:
  - `( )` No review needed


js/components/operatorSignIn/attestation.tsx
```diff
@@ -1,15 +1,18 @@
...
+import { ApiResult } from "../../api";
+import { findEmployeeByBadge } from "../../hooks/useEmployees";
+import { Employee } from "../../models/employee";
 import { useSignInText } from "./text";
 import { ReactElement } from "react";
 
 export const Attestation = ({
   badge,
   onComplete,
+  employees,
```
mathcolo, line 10: I pulled this over to my branch- thanks!

js/components/operatorSignIn/operatorSelection.tsx
```diff
@@ -1,16 +1,40 @@
...
+  employees,
 }: {
   nfcSupported: boolean;
   onOK: (badge: string) => void;
+  employees: ApiResult<Employee[]>;
 }): ReactElement => {
   const inputId = useId();
   const [badgeEntry, setBadgeEntry] = useState<string>("");
 
+  const nfcAbortController = new AbortController();
```
mathcolo, line 20: I forget, can these be reused?

js/components/operatorSignIn/operatorSelection.tsx
```diff
@@ -1,16 +1,40 @@
...
+    if (nfcResult.status === "success" && employees.status === "ok") {
+      const badge = findEmployeeByBadgeSerial(
+        employees.result,
+        nfcResult.data,
+      )?.badge;
+
+      if (badge) {
+        onOK(badge);
+      } else {
+        // handle case where we can't find operator for badge
```
mathcolo, line 33: Is this going to be in a subsequent PR?

js/components/operatorSignIn/operatorSelection.tsx
lemald: Excellent question, I'll have to look into it.

js/components/operatorSignIn/operatorSelection.tsx
lemald: Yes, my thought process is that this would be best implemented once your work is ready, since you're also making additions to handle the state of the sign in flow.

js/components/operatorSignIn/operatorSelection.tsx
lemald: Also I'm realizing that as written this will actually create a new AbortController on every render, which is not what I want. I'm working on a bit of a refactor of the `useNfc` hook now.

mathcolo: OK! I will hang tight for that.

lemald: Okay I refactored things a bit in 10c66ec and 4dcd61b to basically move the creation of the `AbortController` entirely into the `useNfc` hook to make sure it doesn't get reused. The logic is pretty tricky but I think I've got it working right, and given what a change this is I'm definitely going to do some manual testing.

lemald: Okay, as best as I can tell this is working. Incidentally, [this](https://react.dev/learn/state-as-a-snapshot) and other parts of the React docs were helpful reminders about the component lifecycle.

js/hooks/useNfc.ts
```diff
@@ -8,20 +8,36 @@ export type NfcResult =
...
+  abortController: AbortController | null;
+} => {
   const supported = nfcSupported();
 
   const [result, setResult] = useState<NfcResult>({
     status: supported ? "reading" : "nfcUnsupported",
   });
 
+  const [abortController, setAbortController] =
+    useState<AbortController | null>(null);
```
mathcolo, line 22: Should this be a ref? Does its value changing need to result in a re-render?

js/hooks/useNfc.ts
```diff
@@ -31,15 +47,16 @@ export const useNfc = (cancel: AbortController): NfcResult => {
               status: "success",
               data: dehexSerial(event.serialNumber),
             });
-            cancel.abort();
+
+            abortController.abort();
```
mathcolo, line 51: Why call this on success?

js/test/components/home.test.tsx
```diff
@@ -1,6 +1,11 @@
 import { Home } from "../../components/home";
 import { render } from "@testing-library/react";
 
+jest.mock("../../hooks/useEmployees", () => ({
+  __esModule: true,
```
mathcolo, line 5: I was noticing in mocking fun this morning that this line may not be required if there's no default/non-named export involved, so maybe try it without?

js/test/components/operatorSignIn/operatorSignInModal.test.tsx
```diff
@@ -1,14 +1,41 @@
 import { OperatorSignInModal } from "../../../components/operatorSignIn/operatorSignInModal";
+import { nfcSupported } from "../../../util/nfc";
 import { render } from "@testing-library/react";
 import userEvent from "@testing-library/user-event";
 
+jest.mock("../../../util/nfc", () => ({
+  __esModule: true,
```
mathcolo, line 7: Same as above with `__esModule` here (and the rest of the usages in this file)

mathcolo (commented):
A couple more comments. Almost ready to approve!

js/hooks/useNfc.ts
lemald: Since the `abortController` is part of the return value of the hook, we need it to be updated as part of the React state if we want changes to the `abortController` to actually trigger a rerender and therefore a new call to `useNfc` and a new return value for the caller. If it was purely internal to this function I think a ref might make sense (or even just a local variable).

js/hooks/useNfc.ts
lemald: Because on success we want the reader to stop listening for NFC taps. It's a weird edge case, but because of the old bug in Chrome that caused us to not use the abort controller in Glides, if you do a badge tap sign-in and then later on in the flow tap a badge on your phone again with the browser still open, it will make the little vibration indicating that the NFC reader is reading. With this call to `abortController.abort()`, it will stop reading once we get a successful scan and won't start reading again until you start a new sign-in flow and a new copy of the `useNfc` hook is mounted.

js/test/components/home.test.tsx
lemald: Oh, neat, let me take a look.

lemald: 0118b17

js/test/components/operatorSignIn/operatorSignInModal.test.tsx
lemald: 0118b17

lemald: And also 57d8ca2.

js/hooks/useNfc.ts
mathcolo: Would you mind putting that exact explanation in as a comment? Got it.

mathcolo (approved):
I'm good with this! Should mine go first, and then this be modified to hook into that work?

js/hooks/useNfc.ts
lemald: I edited it slightly to better fit the context but here: 23ebe1a

PR #45, opened 2024-07-17
feat: Fit for Duty Check modal interaction

mathcolo:
Asana Task: [Fit for Duty Check modal interaction](https://app.asana.com/0/1206105669438487/1207259534606018/f)

- Success & error screens
- Badge number validation
- Submit to server w/retry

KNOWN ISSUES:
- The ⚠️ TODO `sleep()` on sign_in_controller.ex:20 (there to test the "submitting" state)
- The error screen doesn't show an error message (it's only `console.error`ed), which doesn't match the requirement on the Miro board

Checklist

- Tests:
  - `(x)` Has tests
  - `( )` Doesn't need tests
- Product/Design sign off:
  - `( )` No review needed


js/components/operatorSignIn/attestation.tsx
```diff
@@ -1,37 +1,60 @@
...
+  const [entered, setEntered] = useState<string>("");
+  const ready = entered === badge;
+
   if (employees.status === "loading") {
     return <div>Loading...</div>;
   } else if (employees.status === "error") {
-    return <div>Unable to download employee data. Please try again.</div>;
+    return (
+      <div>
+        Unable to download employee data. Please <a href="/">reload</a> to try
```
lemald, line 27: **issue (non-blocking):** I don't think this is super important right now, but I just want to flag that since the `"/"` path is hardcoded here, this will potentially behave in an undesired way if this component is ever rendered on a page with a different path. Maybe that's easy to change quickly but if not don't worry too much about it.

js/components/operatorSignIn/attestation.tsx
```diff
@@ -40,20 +63,95 @@ export const SignInText = (): ReactElement => {
...
+  } else if (visibleHint === "bad") {
+    hintText = "That badge number doesn't match our records.";
+    hintClass = "text-mbta-red";
+    title = `expected ${JSON.stringify(badge)}, got ${JSON.stringify(
+      signatureText,
+    )}`;
+  } else {
+    hintText = "";
+    hintClass = "leading-[0]";
+  }
```
lemald, line 119: **suggestion:** this may not be a belief shared by everyone, but I suspect this might be a bit easier to read with a `const` assignment and use of the ternary operator. For instance, you could do `const hintText = (visibleHint === "good") ? "Looks good!" : (visibleHint === "bad") ? ...` and then split it across multiple lines, and something similar with `hintClass`.

lemald: Also, sorry, this nitpickiest of nitpicks, but I think either way this would be easier to read with an empty line above and below this section to break things up a bit.

js/components/operatorSignIn/complete.tsx
```diff
@@ -0,0 +1,39 @@
+import { className } from "../../util/dom";
+import { ReactElement } from "react";
+
+export const Success = ({ name }: { name: string }): ReactElement => {
+  return (
+    <div className="text-center text-4xl w-full block mt-5 p-10">
+      ✅<div>{name} signed in successfully</div>
```
lemald, line 7: **question:** the name should also get the `fs-mask` class, right? I know we don't have FullStory yet but I see we've been adding the class in some places, so it would probably be good to be as thorough as possible with it.

lemald: Same question pertains to the name in the `Error` component.

js/test/components/operatorSignIn/operatorSignInModal.test.tsx
```diff
@@ -16,4 +32,42 @@ describe("OperatorSignInModal", () => {
 
     expect(view.queryByText(/fit for duty check/i)).not.toBeInTheDocument();
   });
+
+  test("submits successful attestation to the server", async () => {
+    putMetaData("csrf-token", "TEST-CSRF-TOKEN");
+    (fetch as jest.Mock).mockResolvedValueOnce({
```
lemald, line 38: **suggestion:** I believe this can be rewritten as:
```suggestion
    jest.mocked(fetch).mockResolvedValueOnce({
```
which provides a bit more type safety.

package.json
```diff
@@ -40,6 +41,7 @@
     "@testing-library/user-event": "14.5.2",
     "@types/eslint__js": "8.42.3",
     "@types/jest": "29.5.12",
+    "@types/luxon": "^3.4.2",
```
lemald, line 44: **nitpick:** shouldn't this be hard pinned to 3.4.2 instead?

js/components/operatorSignIn/operatorSignInModal.tsx
```diff
@@ -1,11 +1,73 @@
...
-import { ReactElement, useState } from "react";
+import { DateTime } from "luxon";
+import { ReactElement, useEffect, useState } from "react";
+
+enum CompleteState {
+  SUCCESS,
+  ERROR,
+}
+
+export const submit = (
```
lemald, line 15: **question:** does this actually need to be `export`-ed? I can't find any particular spots where it's used outside of this file.

lib/orbit_web/controllers/sign_in_controller.ex
```diff
@@ -0,0 +1,36 @@
...
+        "method" => method
+      }) do
+    signed_in_at = DateTime.from_unix!(signed_in_at)
+
+    # ⚠️ TODO: remove before merge
+    :timer.sleep(1000)
+
+    %OperatorSignIn{
+      signed_in_employee:
+        Repo.one!(from(e in Employee, where: e.badge_number == ^signed_in_employee_badge)),
```
lemald, line 25: **suggestion:** could we give a more specific response code if no operator for the corresponding badge is found? I think a 404 would be semantically appropriate but I'd have to read a bit more to confirm.

lemald:
Okay @mathcolo I'm done with my first pass of review. I left a lot of little comments, but nothing too major. 👍

lib/orbit_web/controllers/sign_in_controller.ex
mathcolo: Added a test; already works that way
b297dd1

js/components/operatorSignIn/operatorSignInModal.tsx
mathcolo: f1143a7

js/test/components/operatorSignIn/operatorSignInModal.test.tsx
mathcolo: thanks! f1b55ed

js/components/operatorSignIn/complete.tsx
mathcolo: 062b1f7

lib/orbit_web/controllers/sign_in_controller.ex
lemald: Oh, right, I forgot that Ecto + Phoenix does some fancy stuff to turn Ecto exceptions into relevant HTTP response codes that usually make sense. But thanks for adding that test, that's good to have!

js/components/operatorSignIn/attestation.tsx
mathcolo: Nah, you're right- might as well not make things get weird later! 6a78e9d

js/components/operatorSignIn/attestation.tsx
mathcolo: This is tough, because three ternary operators for each variable would be kind of obtuse. I cleaned it up a little (and removed the `JSON.stringify` calls, which I imagine were for surrounding with double quotes?

ce4b293

lemald (approved):

PR #46, opened 2024-07-17
fix: nfc signins get 'nfc' method

mathcolo:
Asana Task: Follow up on [Fit for Duty Check modal interaction](https://app.asana.com/0/1206105669438487/1207259534606018/f)

Title says it all. Merging two earlier PRs meant sign-ins were hardcoded to method "manual"- nfc signins should get "nfc".

Checklist

- Tests:
  - `(x)` Has tests
  - `( )` Doesn't need tests
- Product/Design sign off:
  - `(x)` No review needed

### DIFF
--- a/js/components/operatorSignIn/operatorSelection.tsx
+++ b/js/components/operatorSignIn/operatorSelection.tsx
@@ -54,9 +54,14 @@ export const OperatorSelection = ({
       <button
         disabled={!buttonEnabled}
         onClick={() => {
-          // badgeEntry cannot be null; otherwise the button would be disabled!
-          // eslint-disable-next-line @typescript-eslint/no-non-null-assertion
-          onOK(badgeEntry!);
+          // This should never happen, but throw just in case.
+          if (badgeEntry === null) {
+            throw new Error(
+              "operatorSelection badgeEntry was impossibly null.",
+            );
+          }
+
+          onOK(badgeEntry);
         }}
         className={className([
           "rounded bg-mbta-blue text-gray-200 mt-3 w-1/4 max-w-32 mx-auto",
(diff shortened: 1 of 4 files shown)


js/components/operatorSignIn/operatorSelection.tsx
```diff
@@ -47,13 +48,15 @@ export const OperatorSelection = ({
...
+          setBadgeEntry({ number: evt.target.value, method: "manual" });
         }}
       />
       <button
         disabled={!buttonEnabled}
         onClick={() => {
-          onOK(badgeEntry);
+          // badgeEntry cannot be null; otherwise the button would be disabled!
+          // eslint-disable-next-line @typescript-eslint/no-non-null-assertion
+          onOK(badgeEntry!);
```
lemald, line 59: **suggestion:**  it's a bit more verbose, but rather than doing a non-null assertion, which requires disabling an ESLint rule and loses us a bit of type-safety, this is the sort of case where I'd usually prefer to do something like:

```
if (value !== null) {
  doThing(value) // TypeScript knowns it's not null
} else {
  // report an error, maybe by throwing, `console.error`-ing, or sending an exception to Sentry.
}
```

lemald (changes requested):
Sorry, I should have actually indicated that I'm done with my first pass of the review - please see my comment, everything else looks good.

js/components/operatorSignIn/operatorSelection.tsx
mathcolo: Since this is an (in theory) impossible state, I changed it to throw.

lemald (approved):
👍

PR #51, opened 2024-07-22
feat: View of completed operator sign-ins

mathcolo:
Asana Task: [View of completed operator sign-ins](https://app.asana.com/0/1206105669438487/1207305284303289/f)

image

Checklist

- Tests:
  - `(x)` Has tests
  - `( )` Doesn't need tests
- Product/Design sign off:
  - `( )` No review needed


js/components/operatorSignIn/list.tsx
```diff
@@ -0,0 +1,62 @@
...
+  const employees = useEmployees();
+  const signIns = useSignins(line);
+
+  if (signIns.status === "loading" || employees.status === "loading") {
+    return <>Loading...</>;
+  } else if (signIns.status === "error") {
+    return <>Error retrieving signins: {signIns.error}</>;
+  } else if (employees.status === "error") {
+    return <>Error retrieving employees: {employees.error}</>;
+  }
```
lemald, line 18: **suggestion (non-blocking):** I don't have an immediate idea of how to implement this, but it feels like this could be factored out into some sort of reusable "loading or fetch error" component that could be used for various places where the application needs to wait for API responses.

js/components/operatorSignIn/list.tsx
```diff
@@ -0,0 +1,62 @@
+import { back } from "../../browser";
+import { findEmployeeByBadge, useEmployees } from "../../hooks/useEmployees";
+import { useSignins } from "../../hooks/useSignIns";
+import { fullName } from "../../models/employee";
+import { DateTime } from "luxon";
+import { ReactElement } from "react";
+
+export const List = ({ line }: { line: string }): ReactElement => {
```
lemald, line 8: *suggestion:** this could be a `HeavyRailLine` type or something like that, rather than an arbitrary `string`.

js/models/employee.ts
```diff
@@ -8,6 +8,14 @@ export const Employee = z.object({
   badge_serials: z.array(z.string()),
 });
 
+export const fullName = (emp: Employee | null) => {
```
lemald, line 11: **nitpick:** could this be named something like `displayName` or `fullDisplayName` instead? Usually when I think of "full name" I think of it as including first, middle, and last (with all the usual caveats about names being complicated and not everyone having a name of that form to begin with).

js/models/employee.ts
```diff
@@ -8,6 +8,14 @@ export const Employee = z.object({
...
+  if (emp === null) {
+    return null;
+  }
```
lemald, line 14: **suggestion (non-blocking):** for utility functions like this, I think that I'd generally prefer that the caller handle things like null cases, but I'll admit I haven't thought this through very rigorously.

lib/orbit_web/controllers/sign_in_controller.ex
```diff
@@ -8,6 +8,45 @@ defmodule OrbitWeb.SignInController do
...
+
+    json(
+      conn,
+      %{
+        data:
+          Repo.all(
+            from(si in OperatorSignIn,
+              where:
+                ^start_datetime < si.signed_in_at and
+                  si.signed_in_at < ^end_datetime and
```
lemald, line 32: **nitpick:** I think that one or both of these inequalities should be inclusive, just in case we ever hit the edge case where someone is signed in during the second in which the service date rolls over.

lib/util/time.ex
```diff
@@ -0,0 +1,40 @@
...
+  @type timestamp :: integer()
+
+  @spec current_time :: timestamp
+  def current_time do
+    System.system_time(:second)
+  end
+
+  @spec current_datetime :: DateTime.t()
+  def current_datetime do
+    DateTime.now!("America/New_York")
```
lemald, line 14: **suggestion:** this, and the other places we reference the local time zone, could go in an application configuration variable (like `config/config.exs`).

test/orbit_web/controllers/sign_in_controller_test.exs
```diff
@@ -7,6 +7,74 @@ defmodule OrbitWeb.SignInControllerTest do
 
   import Ecto.Query
 
+  describe "index" do
+    @tag :authenticated
```
lemald, line 11: **suggestion:** it's probably worth having a test (both for `index` and also `submit`) making sure that unauthenticated connections receive the appropriate response code, and don't get any sensitive data.

lemald (approved):
Okay, I'm approving because overall I don't think there's anything _truly_ blocking about any of the things I flagged, but please take a look and let me know what you think of my feedback. 👍

lib/orbit_web/controllers/sign_in_controller.ex
mathcolo: I'd prefer not to change it. It's unambiguous that < 3:00 is the previous service day, and it's easily explainable that way. Further, sign-ins are anecdotally rarely < 4:00

lib/util/time.ex
mathcolo: I'll add it to the config. But I'm afraid of removing it from `current_datetime` and other time functions, since the consequences of timezone issues are so severe...

lib/orbit_web/controllers/sign_in_controller.ex
lemald: Right, < 3:00 is the previous service day, but exactly 3:00 is the current service day, and with the code as-written 3:00 on the dot doesn't fall into either service date. I agree that it's highly unlikely to ever be a problem in practice, but I think theoretically at least it's incorrect.

lib/util/time.ex
lemald: I'm confused - we wouldn't be removing references to time zones in general, it's just that we would point to the configuration for the particular time zone in question.

lib/orbit_web/controllers/sign_in_controller.ex
mathcolo: OH! Gotcha. Yep, I agree. Will fix 👍

js/components/operatorSignIn/list.tsx
mathcolo: ec7c168

js/models/employee.ts
mathcolo: c959b6e

js/models/employee.ts
mathcolo: (same as above) c959b6e

lib/orbit_web/controllers/sign_in_controller.ex
mathcolo: 6e54d41

test/orbit_web/controllers/sign_in_controller_test.exs
mathcolo: 7e57dd0

mathcolo: (and 53bf3b0)

lib/util/time.ex
mathcolo: Oh, I thought this was a global timezone override or something, akin to the TZ environment variable in some runtimes. Yeah, I'll change this.

lemald: Ah, yeah, I don't mean some kind of global override, just a configuration value that we could directly reference whenever we need the name of a time zone for invoking some `DateTime` or `Timex` function.

mathcolo: 2b6e3ff

js/components/operatorSignIn/list.tsx
mathcolo: Yeah, agree. I'll file a task- want to get this merged in the immediate future.

lemald: Sounds good!

lemald (approved):
Looks good to go!

PR #54, opened 2024-07-23
fix: operators without rfid serials weren't included in API results

lemald:
Asana Task: [🐞 orbit: some operators not being returned in API results](https://app.asana.com/0/1200273269966439/1207834935765422/f)

I'm not sure if this fixes all of the exact issues, but it's definitely a case that would cause this problem.

I took inspiration from the second example under [here](https://hexdocs.pm/ecto/Ecto.Query.html#join/5-keywords-examples), the one querying on `Post` and joining to `comments`, since that was a similar case to our own.

Checklist

- Tests:
  - `(X)` Has tests
  - `( )` Doesn't need tests
- Product/Design sign off:
  - `(X)` No review needed

### DIFF
--- a/lib/orbit_web/controllers/employees_controller.ex
+++ b/lib/orbit_web/controllers/employees_controller.ex
@@ -11,8 +11,7 @@ defmodule OrbitWeb.EmployeesController do
       Enum.map(
         Repo.all(
           from e in Employee,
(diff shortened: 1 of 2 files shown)


mathcolo (approved):
Looks like Ecto's `join` keyword means inner join by default. Tricky, but nice fix! Works for me locally.

PR #58, opened 2024-07-24
Frontend Sentry

mathcolo:
Asana Task: [📐 Orbit install Sentry on frontend](https://app.asana.com/0/1206105669438487/1207620537833336/f)

- Add Sentry
- Add error boundary
- Capture exceptions across the board

Checklist

- Tests:
  - `(x)` Has tests **(Has one for metadata. It's otherwise mostly configuration and CSS, so not seeing a huge additional need.)**
  - `( )` Doesn't need tests
- Product/Design sign off:
  - `( )` No review needed

### DIFF
--- a/config/runtime.exs
+++ b/config/runtime.exs
@@ -9,6 +9,8 @@ import Config
 if config_env() == :prod do
   maybe_ipv6 = if System.get_env("ECTO_IPV6") in ~w(true 1), do: [:inet6], else: []
 
+  config :orbit, environment: System.get_env("ENVIRONMENT")
+
   config :orbit, Orbit.Repo,
     socket_options: maybe_ipv6,
     username: System.get_env("DATABASE_USERNAME"),
@@ -87,6 +89,5 @@ if config_env() == :prod do
     secret_key_base: secret_key_base
 
   config :sentry,
-    dsn: System.get_env("SENTRY_DSN"),
-    environment_name: System.get_env("ENVIRONMENT")
+    dsn: System.get_env("SENTRY_DSN")
 end
--- a/config/runtime.exs
+++ b/config/runtime.exs
@@ -89,5 +89,6 @@ if config_env() == :prod do
     secret_key_base: secret_key_base
 
   config :sentry,
-    dsn: System.get_env("SENTRY_DSN")
+    dsn: System.get_env("SENTRY_DSN"),
+    environment_name: System.get_env("ENVIRONMENT")
 end
--- a/js/app.tsx
+++ b/js/app.tsx
@@ -1,3 +1,4 @@
+import "./sentry.js";
 import "./phoenix.js";
 import { App } from "./components/app.js";
 import { createRoot } from "react-dom/client";
--- a/js/app.tsx
+++ b/js/app.tsx
@@ -1,4 +1,5 @@
-import "./sentry.js";
+// runs initialization during import
+import "./sentry";
 import "./phoenix.js";
 import { App } from "./components/app.js";
 import { createRoot } from "react-dom/client";
--- a/js/components/app.tsx
+++ b/js/components/app.tsx
@@ -14,7 +14,7 @@ const ErrorBoundary = (): ReactElement => {
   const error = useRouteError() as Error;
   useEffect(() => {
     captureException(error);
-  });
+  }, [error]);
   return (
     <main className="container mx-auto">
       <p>Sorry, something went wrong. It&apos;s not you, it&apos;s us.</p>
--- a/js/components/app.tsx
+++ b/js/components/app.tsx
@@ -1,21 +1,62 @@
+import { reload } from "../browser";
 import { Help } from "./help";
 import { Home } from "./home";
 import { List } from "./operatorSignIn/list";
-import { ReactElement } from "react";
-import { createBrowserRouter, RouterProvider } from "react-router-dom";
+import { captureException } from "@sentry/react";
+import { ReactElement, useEffect } from "react";
+import {
+  createBrowserRouter,
+  RouterProvider,
+  useRouteError,
+} from "react-router-dom";
+
+const ErrorBoundary = (): ReactElement => {
+  const error = useRouteError() as Error;
+  useEffect(() => {
+    captureException(error);
+  });
+  return (
+    <main className="container mx-auto">
+      <p>Sorry, something went wrong. It&apos;s not you, it&apos;s us.</p>
+      <p>Things you could try:</p>
+      <ul className="list-inside list-disc mb-10">
+        <li>
+          <button type="button" className="underline" onClick={reload}>
+            Refresh the page
+          </button>
+          .
+        </li>
+        <li>
+          <a href="/">Go back to the Orbit home page</a>.
+        </li>
+        <li>Try again in a few minutes.</li>
+        <li>Call the IT help desk. x5761</li>
+      </ul>
+      <p>
+        Error details: {error.name} {error.message}
+      </p>
+      <pre className="bg-gray-200">{error.stack}</pre>
+    </main>
+  );
+};
 
 const router = createBrowserRouter([
   {
-    path: "/",
-    element: <Home />,
-  },
-  {
-    path: "/list",
-    element: <List line="blue" />,
-  },
-  {
-    path: "/help",
-    element: <Help />,
+    errorElement: <ErrorBoundary />,
+    children: [
+      {
+        path: "/",
+        element: <Home />,
+      },
+      {
+        path: "/list",
+        element: <List line="blue" />,
+      },
+      {
+        path: "/help",
+        element: <Help />,
+      },
+    ],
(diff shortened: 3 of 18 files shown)


package.json
```diff
@@ -14,6 +14,7 @@
   },
   "dependencies": {
     "@restart/ui": "1.6.9",
+    "@sentry/react": "^8.19.0",
```
lemald, line 17: **nitpick:**
```suggestion
    "@sentry/react": "8.19.0",
```

mathcolo: This I may never learn. thanks

js/components/app.tsx
```diff
@@ -1,21 +1,62 @@
...
+import { captureException } from "@sentry/react";
+import { ReactElement, useEffect } from "react";
+import {
+  createBrowserRouter,
+  RouterProvider,
+  useRouteError,
+} from "react-router-dom";
+
+const ErrorBoundary = (): ReactElement => {
+  const error = useRouteError() as Error;
```
lemald, line 14: **suggestion:** I see that this just returns `unknown` in the types provided by the package, which is unfortunate. It might be possible to gate the `captureException` call behind an `if (error instanceof Error)` condition or something, but then we probably also want to do something if it's not an instance of `Error` because that's itself a weird case we want visibility into.

js/sentry.ts
```diff
@@ -0,0 +1,19 @@
...
+  Sentry.init({
+    dsn: sentryDsn,
+    environment: sentryEnvironment,
+    release: release ?? undefined,
+  });
+
+  if (userEmail !== null) {
+    Sentry.setUser({ email: userEmail });
+  }
+}
```
lemald, line 19: **suggestion:** rather than having this code run directly at the top level on import, how about wrapping it in an exported `initializeSentry` function and calling it explicitly from `js/app.tsx`? I personally find that more explicit / easier to reason about.

test/orbit_web/controllers/react_app_controller_test.exs
```diff
@@ -6,7 +6,9 @@ defmodule OrbitWeb.ReactAppControllerTest do
     conn = get(conn, ~p"/")
     response = html_response(conn, 200)
     assert response =~ "Orbit"
-    # sets release
+
+    # sets metadata
     assert response =~ "<meta name=\"release\" content=\"test\">"
+    assert response =~ "<meta name=\"userEmail\" content=\"user@example.com\">"
```
lemald, line 12: **suggestion:** can you also test that the relevant Sentry `<meta/>` tags are present?

lemald (changes requested):
Done with my first pass, there are just a handful of things that need addressing.

js/components/app.tsx
```diff
@@ -1,21 +1,62 @@
...
+  useEffect(() => {
+    captureException(error);
+  });
```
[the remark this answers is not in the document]
mathcolo: thanks! db4d849

js/sentry.ts
mathcolo: Interestingly they actually recommend this, probably to avoid other global scope code from beating it. https://docs.sentry.io/platforms/javascript/guides/react/

test/orbit_web/controllers/react_app_controller_test.exs
mathcolo: I actually tried, but putting test Sentry configuration in `test.exs` causes backend Sentry to try to load and connect to those fake resources that don't exist. I could mock `get_env` itself here, but that seems kinda weird. Thoughts?

lemald: Would something like [this macro](https://github.com/mbta/skate/blob/main/test/support/helpers.ex#L4) that we've copy-pasted to a few repos help? It basically sets the variable while also setting up an `on_exit` to clean up the value so that it doesn't interfere with other tests.

js/sentry.ts
lemald: Ah, okay - if it's recommended by the docs then I'm fine with it.

test/orbit_web/controllers/react_app_controller_test.exs
mathcolo: Ooh! Like. I'll give it a shot.

js/app.tsx
```diff
@@ -1,3 +1,4 @@
+import "./sentry.js";
```
skyqrose, line 1: Right now all the side effects happen during the import because they're top level in `sentry.ts`, which makes it hard to track all the things that happen during initialization.

Could you wrap all the code in that file in a `export const init = () => {}`, and then call the init from this file?

(Also, this is written `./sentry.js`, when I think `./sentry` should be fine. I'm surprised it worked and also surprised eslint didn't catch that.)

js/components/app.tsx
skyqrose: `captureException(error)` worked elsewhere when it was `unknown`, so I think the only unsafe use here is when displaying `error.name/message/stack` in the UI. So maybe the `if (error instanceof Error)` check can go down there, and we show something less useful if we can't figure out what type `error` is.

lib/orbit_web/controllers/react_app_controller.ex
```diff
@@ -4,6 +4,10 @@ defmodule OrbitWeb.ReactAppController do
   def home(conn, _params) do
     # The home page is often custom made,
     # so skip the default app layout.
-    render(conn, :react_app, layout: false)
+    render(conn, :react_app,
+      layout: false,
+      sentry_dsn: Application.get_env(:sentry, :dsn),
+      sentry_environment: Application.get_env(:sentry, :environment_name)
```
skyqrose, line 10: In Glides, we end up referring to `sentry_release` in a bunch of non-sentry places, because sentry was the first use of a release name. But it's not really a sentry-specific concept, so the name is a bit misleading.

I'd like to try to fix that problem in orbit. The environment is pretty generic concept. Can we add a `config :orbit, environment: System.get_env("ENVIRONMENT")` in `runtime.exs` so that we can refer to this everywhere as `environment: Application.get_env(:orbit, :environment)` and not tie it to the sentry name?

test/orbit_web/controllers/react_app_controller_test.exs
skyqrose: If the meta tag is based on `Application.get_env(:orbit, :environment)`, (from my other comment) then the meta tag could be set even without turning on sentry.

js/app.tsx
mathcolo: See my comment here: https://github.com/mbta/orbit/pull/58#discussion_r1691718215

skyqrose: whaaat

An alternative could be to add a comment and add an empty line to separate it from all the well-behaved imports:

```suggestion
// runs initialization during import
import "./sentry";

```

lib/orbit_web/controllers/react_app_controller.ex
mathcolo: yep! fixed in above.

test/orbit_web/controllers/react_app_controller_test.exs
mathcolo: fixed in d463f99

js/components/app.tsx
mathcolo: 👍 let me know what you both think about 5c0c72 ?

lemald: I like it!

lemald (approved):
Looks like everything I personally flagged has been addressed.

config/runtime.exs
```diff
@@ -87,6 +89,5 @@ if config_env() == :prod do
     secret_key_base: secret_key_base
 
   config :sentry,
-    dsn: System.get_env("SENTRY_DSN"),
-    environment_name: System.get_env("ENVIRONMENT")
```
skyqrose, line 91: I think this config value is still needed for sentry's use on the backend. So it can still reference `System.get_env("ENVIRONMENT")`. Sorry for being unclear. We should have both `config :sentry, evnironment_name` AND `config, :orbit, :environment`, one for sentry (backend) to use, and one for everything else to use.

js/test/components/app.test.tsx
```diff
@@ -0,0 +1,20 @@
+import { App } from "../../components/app";
+import { render } from "@testing-library/react";
+
+describe("App", () => {
+  test("error boundary renders on error", () => {
+    jest.mock("../../components/operatorSignIn/operatorSignInModal", () => ({
+      OperatorSignInModal: () => {
+        throw new Error(
+          "Ah, Houston, we've had a problem. We've had a Main B Bus Undervolt.",
```
skyqrose, line 9: 🚀😱

config/runtime.exs
mathcolo: Ah! gotcha. I should have tested in sandbox like I was doing yesterday. Will fix-

PR #59, opened 2024-07-24
Basic user permissions

lemald:
Asana Task: [Set up user access / permissions](https://app.asana.com/0/1200273269966439/1207259534605993/f)

This is just the super basic database-level means of storing user permissions. I have tentatively added operator sign in as the first permission, figuring that we'll definitely need a permission for that. New permissions can be added in subsequent migrations via `ALTER TYPE` statements, though note that with enums in Postgres you can't remove enum values once they've been added.

As far as helped functions and CRUD operations, if I add those it will be in a followup PR, but the out-of-the-box Ecto schema stuff already makes it easy to fetch and update these values, so I'm divided. I think where it will make sense is once we're dealing with additional permissions from Keycloak, in which case it will make sense to have a combined `user_permissions_for_conn` function or something like that.

One additional thing I wanted to add, since I love database constraints, is a constraint that the elements of the array be unique, so that we don't have the same permission repeated multiple times for a single user. Unfortunately, it looks like that would be a "check constraint," and check constraints can't use things like subqueries or aggregate functions, which are the tools I would need in order to check for duplicate elements. Maybe it's possible to write a constraint that wouldn't need to use one of those features, but if so it's probably more complicated than it's worth.

Checklist

- Tests:
  - `(X)` Has tests
  - `( )` Doesn't need tests
- Product/Design sign off:
  - `(X)` No review needed

### DIFF
--- a/lib/orbit/authentication/user.ex
+++ b/lib/orbit/authentication/user.ex
@@ -1,25 +1,29 @@
 defmodule Orbit.Authentication.User do
+  alias Orbit.Authentication.UserPermission
   use Ecto.Schema
   import Ecto.Changeset
 
   @type t :: %__MODULE__{
           id: integer(),
           email: String.t(),
+          permissions: [UserPermission.t()],
           inserted_at: DateTime.t(),
           updated_at: DateTime.t()
         }
 
   schema "users" do
     field(:email, :string)
+    field(:permissions, {:array, UserPermission})
     timestamps()
   end
 
   def create_changeset(struct, params \\ %{}) do
     struct
     |> cast(params, [
-      :email
+      :email,
+      :permissions
     ])
-    |> validate_required([:email])
+    |> validate_required([:email, :permissions])
     |> unique_constraint(
       :users,
       name: :users_email_index
--- a/lib/orbit/authentication/user.ex
+++ b/lib/orbit/authentication/user.ex
@@ -17,13 +17,20 @@ defmodule Orbit.Authentication.User do
     timestamps()
   end
 
-  def create_changeset(struct, params \\ %{}) do
+  def changeset(struct, params \\ %{}) do
     struct
     |> cast(params, [
       :email,
       :permissions
     ])
     |> validate_required([:email, :permissions])
+    |> validate_change(:permissions, fn :permissions, permissions ->
+      if permissions |> Enum.uniq() |> length() == length(permissions) do
+        []
+      else
+        [permissions: "cannot contain duplicate values"]
+      end
+    end)
     |> unique_constraint(
       :users,
       name: :users_email_index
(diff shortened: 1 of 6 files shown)


priv/repo/migrations/20240724173817_add_user_permissions_column.exs
```diff
@@ -0,0 +1,14 @@
...
+  use Ecto.Migration
+
+  def change do
+    execute(
+      "CREATE TYPE user_permission AS ENUM ('operator_sign_in')",
+      "DROP TYPE user_permission"
+    )
+
+    alter table("users") do
+      add :permissions, {:array, :user_permission}
```
mathcolo, line 11: Is this NOT NULL (in the sql sense) by default? i.e. is NULL vs empty array something you're aware of and is handled?

mathcolo (changes requested):
Just want a little bit of certainty on the null/empty array question before approving.

lib/orbit/authentication/user.ex
```diff
@@ -1,25 +1,29 @@
...
           updated_at: DateTime.t()
         }
 
   schema "users" do
     field(:email, :string)
+    field(:permissions, {:array, UserPermission})
     timestamps()
   end
 
   def create_changeset(struct, params \\ %{}) do
```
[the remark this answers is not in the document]
lemald: Ah, good point! It's not quite as nice as enforcing it at the DB level but still good, I'll go ahead and add something like that.

priv/repo/migrations/20240724173817_add_user_permissions_column.exs
lemald: Good catch - I thought it usually would default to not null, but this is adding a new column so the default behavior is probably different. I made it non-nullable and added a default value of an empty array in [].

lib/orbit/authentication/user.ex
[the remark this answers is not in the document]
lemald: Okay, I gave it a try - check out 062e03d.

priv/repo/migrations/20240724173817_add_user_permissions_column.exs
lemald: 3087d2a

lib/orbit/authentication/user.ex
```diff
@@ -1,25 +1,36 @@
...
     struct
     |> cast(params, [
-      :email
+      :email,
+      :permissions
     ])
-    |> validate_required([:email])
+    |> validate_required([:email, :permissions])
+    |> validate_change(:permissions, fn :permissions, permissions ->
+      if permissions |> Enum.uniq() |> length() == length(permissions) do
```
mathcolo, line 28: Since elixir lists are compared by element, can this instead be `if Enum.uniq(permissions) != permissions`?

lemald: Hmm, so my one fear is that [the docs on `Enum.uniq/1`](https://hexdocs.pm/elixir/Enum.html#uniq/1) don't actually say that it preserves the order of elements. It probably does in practice, but I don't want to rely on undocumented behavior like that. In contrast, [`Enum.group_by/3`s documentation](https://hexdocs.pm/elixir/Enum.html#group_by/3) does specify that it preserves the relative order of elements in the resulting lists within each map value.

mathcolo (approved):

PR #62, opened 2024-07-25
Landing Page

skyqrose:
Asana Task: [Orbit landing page](https://app.asana.com/0/1200689710863995/1207769401354149/f)

Screen Shot 2024-07-25 at 15 18 11

The date picker works, but the button to do the export doesn't do anything yet.

The search input also doesn't exist yet.

I've made no attempt to use any sort of reusable components for the colors and buttons, cuz I assume that'll come later.

Checklist

- Tests:
  - `(x)` Has tests
  - `( )` Doesn't need tests
  - `(x)` Tests deferred (with justification)
    - test for opening the modal
    - There's funcationality for the behavior of the input that makes more sense to test once that input actually controls anything.
    - The logout button is untested, but I wasn't sure whether there'd be a good way to test it, since it's just a link. Should we have a top level `app.test.tsx`?
- Product/Design sign off:
  - `( )` No review needed


js/components/home.tsx
```diff
@@ -1,24 +1,73 @@
...
-      <OperatorSignInModal />
-    </>
+      <div className="flex justify-between gap-2 items-end mb-3">
+        <label className="flex flex-col">
+          <span className="text-sm">Service Date</span>
+          <input
+            type="date"
+            value={selectedDate}
+            onChange={(evt) => {
+              console.log(evt.target.value);
```
mathcolo, line 35: nit: superfluous console log

js/test/setupTest.ts
```diff
@@ -1,5 +1,11 @@
 import "@testing-library/jest-dom";
+import { neverPromise } from "./helpers/promiseWithResolvers";
 // add all jest-extended matchers
 import * as jestExtendedMatchers from "jest-extended";
 
 expect.extend(jestExtendedMatchers);
+
+jest.mock("../browser", () => ({
+  fetch: jest.fn(() => neverPromise()),
```
mathcolo, line 9: Could a non-browser runtime appear to hang if it refuses to exit until this never-fulfilling promise finishes? Or rely on the testing framework's timeout?

IDK, just seems like an anti-pattern.

js/components/home.tsx
```diff
@@ -1,24 +1,73 @@
...
+              setSelectedDate(evt.target.value);
+            }}
+            className="rounded"
+          />
+        </label>
+        <button
+          className="rounded border border-[#929292] hover:bg-[#F6F6F6] px-2 py-1 disabled:text-[#929292]"
```
lemald, line 42: **nitpick:** #929292 at least exists in out color palette defined in the Tailwind config and doesn't need to be directly given as a hex value here. I don't see #f6f6f6 but we should probably add that as well.

lemald: Same goes for a handful of other hex values throughout the file.

js/test/setupTest.ts
lemald: I don't think that other runtimes like the one in which our Jest tests run would generally block on a promise until it resolves / rejects, otherwise you couldn't interleave other actions between the creation of the promise and its fulfillment / rejection.

skyqrose: This promise won't control when the test exits (unless you put `await fetch()` in your test). It won't cause hitting the jest timeout (and I just verified this by playing with the timeout setting).

This will mean that any components that make api calls (unless they mock fetch separately) will show loading throughout the test, which seemed the most stable and predictable behavior. Other options are:
- not giving an implementation (it returns undefined, and `fetch().then()` crashes)
- always resolving (no good choice for what to return)
- always rejecting (I think always showing loading is a better default behavior for a test than always showing an error, given all the ways we might respond to an error)

Any test where the result of a fetch matters (or where it matters whether it resolves) should pass their own mock implementation that returns a promise that the test controls.

lemald: For instance, in [this test](https://github.com/mbta/orbit/blob/main/js/test/hooks/useNfc.test.ts#L23) the promise does eventually get rejected, but other things happen between the creation of the promise and the call to `reject`, and those aren't blocked.

js/components/home.tsx
lemald: Looks like this has been addressed, thanks!

js/test/setupTest.ts
mathcolo: OK! My concern is assuaged 👍

lemald (approved):
Looks good from my perspective.

PR #94, opened 2024-08-05
Rename and Restyle some Colors

lemald:
Asana Task: no ticket, based on ad-hoc work from design.

Most of the changes are no-ops from an actual visual styling perspective, but I spoke with design about the ones that weren't, namely changing the old gray-200 and the green color - these changes were intended to remap to the new colors.

Tailwind seems to heavily encourage just using the swatch-based naming system of either color-number or just color, so I'm sticking with that rather than also trying to shoehorn in semantic names, even if that might be my personal preference. Semantic names would also get very tricky once we try to do a dark mode.

Checklist

- Tests:
  - `( )` Has tests
  - `(X)` Doesn't need tests
- Product/Design sign off:
  - `( )` No review needed


tailwind.config.ts
```diff
@@ -20,25 +20,23 @@ export default {
   darkMode: ["class", '[data-mode="dark"]'],
   theme: {
     colors: {
-      "mbta-red": "#DA291C",
-      "mbta-orange": "#ED8B00",
-      "mbta-blue": "#003DA5",
```
skyqrose, line 25: I read the docs on this: https://tailwindcss.com/docs/customizing-colors#naming-your-colors

> we recommend sticking to the default naming convention for most projects, and only using abstract names if you have a really good reason.
> We think this is the best choice for most projects, and have found it easier to maintain than using abstract names like primary or danger.

It doesn't specifically call out this case, but I think maybe these brand colors could be an exception. There are now two different red colors, and I don't know which one is the Red Line red. If we use an inconsistent success-green because we think that green-200 looks better in a place than green-300, then that's fine, so we shouldn't get hung up on "success green" or "default background color", but if we use the wrong Red Line red, that's wrong. I think it still makes sense to have these colors labeled as brand colors, even if the rest of the system is the tailwind recommended numbered way. (Maybe these colors can be usable in both ways.)

skyqrose (approved):
One comment (I'm sad about losing brand color names), tho 👍 approved with or without that.

tailwind.config.ts
lemald: Yes, that's a good point. I went ahead and added the line colors back in in 0861952, though I believe at this point all of the existing uses of the colors in the application are for non-brand purposes, so I've left those as-is.

PR #98, opened 2024-08-20
Admin UI to make fake employees and rfid

skyqrose:
Asana Task: [📐 Orbit: allow manually associating a badge serial with an operator for testing](https://app.asana.com/0/1200689710863995/1207890347662644/f)

Also includes:
- Making employees. It wasn't part of the ticket, but I needed to have someone to attach the new badge to.
- Checkbox on the dev login page to make joining the admin group easier (since you have to do it every time you log in).
- Config to allow making test data everywhere except in prod.

TODO on Thursday:
- [x] Infrastructure PR to set the new `ALLOW_TEST_DATA=yes` env var on staging and sandbox.
  - Infra PRs for adding the env var are now open:
  - https://github.com/mbta/terraform_modules/pull/42
  - https://github.com/mbta/devops/pull/2103
- [x] Write tests. Everything here is complete and ready for review, but I haven't written unit tests yet. I figured I'd get the review process started by opening a draft PR.

Edit: Everything's done, ready for review.

Screenshots:
> Screenshot 2024-08-20 at 17 24 48

> Screenshot 2024-08-20 at 17 25 03

> Screenshot 2024-08-20 at 17 25 11

Checklist

- Tests:
  - `( )` Has tests
  - `( )` Doesn't need tests
- Product/Design sign off:
  - `( )` No review needed

### DIFF
--- a/lib/orbit_web/auth/strategy/fake_oidcc.ex
+++ b/lib/orbit_web/auth/strategy/fake_oidcc.ex
@@ -19,7 +19,7 @@ defmodule OrbitWeb.Auth.Strategy.FakeOidcc do
               </label>
             </div>
             <div>
-            <label >
+            <label>
               <input type="checkbox" name="groups[]" value="orbit-admin">
               orbit-admin group
             </label>
--- a/lib/orbit_web/auth/strategy/fake_oidcc.ex
+++ b/lib/orbit_web/auth/strategy/fake_oidcc.ex
@@ -12,8 +12,21 @@ defmodule OrbitWeb.Auth.Strategy.FakeOidcc do
         <html>
           <h1>Fake Keycloak/Oidcc</h1>
           <form action="/auth/keycloak/callback">
-            <label>Email: <input type="email" name="email" value="user@example.com"></label>
-            <input type="submit" value="Log In">
+            <div>
+              <label>
+                Email:
+                <input type="email" name="email" value="user@example.com">
+              </label>
+            </div>
+            <div>
+            <label >
+              <input type="checkbox" name="groups[]" value="orbit-admin">
+              orbit-admin group
+            </label>
+            </div>
+            <div>
+              <input type="submit" value="Log In">
+            </div>
           </form>
         </html>
       )
--- a/lib/orbit_web/controllers/admin/admin_controller.ex
+++ b/lib/orbit_web/controllers/admin/admin_controller.ex
@@ -1,15 +1,11 @@
-defmodule OrbitWeb.AdminController do
+defmodule OrbitWeb.Admin.AdminController do
   use OrbitWeb, :controller
 
   plug(OrbitWeb.Plugs.RequireAdmin)
 
   @spec get_rfid(Plug.Conn.t(), map()) :: Plug.Conn.t()
   def get_rfid(conn, _params) do
-    conn
-    |> html(~s"""
-    <h1>admin/rfid</h1>
-    <p>Manually Create RFID Badge Serial</p>
-    """)
+    render(conn, :rfid, layout: false)
   end
 
   @spec post_rfid(Plug.Conn.t(), map()) :: Plug.Conn.t()
--- a/lib/orbit_web/controllers/admin/admin_controller.ex
+++ b/lib/orbit_web/controllers/admin/admin_controller.ex
@@ -8,6 +8,14 @@ defmodule OrbitWeb.Admin.AdminController do
 
   plug(OrbitWeb.Plugs.RequireAdmin)
 
+  plug(
+    OrbitWeb.Plugs.RequireAllowTestData
+    when action in [
+           :post_employee,
+           :post_rfid
+         ]
+  )
+
   @spec get_employee(Plug.Conn.t(), map()) :: Plug.Conn.t()
   def get_employee(conn, _params) do
     render(conn, :employee, layout: false)
--- a/lib/orbit_web/controllers/admin/admin_controller.ex
+++ b/lib/orbit_web/controllers/admin/admin_controller.ex
@@ -2,8 +2,8 @@ defmodule OrbitWeb.Admin.AdminController do
   use OrbitWeb, :controller
   import Ecto.Query
 
-  alias Orbit.Employee
   alias Orbit.BadgeSerial
+  alias Orbit.Employee
   alias Orbit.Repo
 
   plug(OrbitWeb.Plugs.RequireAdmin)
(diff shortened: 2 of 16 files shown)


lib/orbit_web/controllers/admin/admin_controller.ex
```diff
@@ -0,0 +1,114 @@
...
+             badge_serial != "" do
+    employee = Repo.one(from e in Employee, where: e.badge_number == ^badge_number)
+
+    if employee != nil do
+      %BadgeSerial{
+        employee: employee,
+        badge_serial: badge_serial
+      }
+      |> BadgeSerial.changeset()
+      |> Repo.insert!()
```
skyqrose, line 99: I didn't do any check that the badge serial is unique, since the database doesn't have that constraint and `import/rfid.ex` doesn't do that check.

Should we enforce uniqueness? How would it behave if there are multiple rows for the same badge serial?

lib/orbit_web/controllers/admin/admin_controller.ex
```diff
@@ -0,0 +1,114 @@
...
+  @spec post_rfid(Plug.Conn.t(), map()) :: Plug.Conn.t()
+  def post_rfid(conn, %{
+        "badge_number" => badge_number,
+        "badge_serial" => badge_serial
+      })
+      when is_binary(badge_number) and
+             badge_number != "" and
+             is_binary(badge_serial) and
+             badge_serial != "" do
+    employee = Repo.one(from e in Employee, where: e.badge_number == ^badge_number)
```
skyqrose, line 91: This doesn't do the `X` prefix to the badge number that `post_employee` does, which means you can make a fake badge serial for a real employee. Should it?

lib/orbit_web/controllers/admin/rfid.html.heex
```diff
@@ -0,0 +1,14 @@
+<h1>admin/rfid</h1>
+<p>Manually Create RFID Badge Serial</p>
+
+<.form for={%{}} action={~p"/admin/rfid"} method="post">
+  <label class="block">
```
skyqrose, line 5: These forms are in a bit of an awkward CSS zone, in that we don't have designs for them and shouldn't care too much, but also tailwind is loaded, which means that default browser styles are disabled. So I ended up doing just a little bit of awkward styles for this and the button, and there's missing styles for stuff that's normally styled like required inputs.

Any ideas on what a better way to handle this would be? Should we fully style this? (On Glides, the admin pages benefit from default styles that orbit doesn't have.) Is there a way to remove the stylesheet from this page so we can use the default browser styles?

lib/orbit_web/controllers/admin/admin_controller.ex
lemald: I don't immediately see any reason why it shouldn't be allowed given that this functionality is only available on non-prod environments.

lib/orbit_web/controllers/admin/admin_controller.ex
lemald: Hmm, good question. I think in any event how to handle it is probably out of scope for this particular PR so I wouldn't worry too much about it, but we should address it somewhere.

lib/orbit_web/controllers/admin/rfid.html.heex
lemald: My initial reaction is that the screenshots don't look perfect but they're fine as-is, though I'm sure there's _some_ way to remove the Tailwind CSS if we really cared to do so.

lemald:
One AC from the ticket is still missing:

> Also have a delete button that allows undoing the association between the entered badge number and badge serial number

I know Glides doesn't have this functionality, but I think it would be good to have if you want to do some testing with a given physical badge for a given operator, then un-associate that physical badge from the operator and associate it with another operator. This is also at least tangentially-related to the question about the uniqueness of badge serials in the mapping table.

lemald:
Anyhow, I have a couple of minor comments and one [major comment](https://github.com/mbta/orbit/pull/98#issuecomment-2302344966), but overall this looks pretty good, I'm just going to hold off an final approval until tests and stuff are in. 👍

lemald:
> We'll need to do either the delete button or the uniqueness, in order to be able to reliably override an existing badge number. I'll one or both of those as a followup PR.

I suppose in theory the delete button is slightly more flexible in that it allows you to dis-associate a badge completely so that it's no longer assigned to any operator, which might be a useful manual test case to be able to do.

lib/orbit_web/auth/strategy/fake_oidcc.ex
```diff
@@ -12,8 +12,21 @@ defmodule OrbitWeb.Auth.Strategy.FakeOidcc do
...
-            <label>Email: <input type="email" name="email" value="user@example.com"></label>
-            <input type="submit" value="Log In">
+            <div>
+              <label>
+                Email:
+                <input type="email" name="email" value="user@example.com">
+              </label>
+            </div>
+            <div>
+            <label >
```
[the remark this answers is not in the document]
skyqrose: [1e779aa](https://github.com/mbta/orbit/pull/98/commits/1e779aa6e6c043c890b1a23d8ba72f7e1e05ed33)

test/orbit_web/controllers/admin_controller_test.exs
```diff
@@ -0,0 +1,93 @@
...
+      conn = post(conn, ~p"/admin/employee", %{})
+      assert response(conn, :method_not_allowed)
+    end
+  end
+
+  describe "employees" do
+    test "get loads form", %{conn: conn} do
+      conn = get(conn, ~p"/admin/employee")
+      assert response = html_response(conn, 200)
+      assert response =~ "<button"
```
lemald, line 37: **nitpick:** doesn't really change the correctness of the test but the unbalanced brackets are bothering me 😅 
```suggestion
      assert response =~ "<button>"
```

lemald (approved):
A couple of nitpicks and removing / updating the badge serial mapping still needs to go in a followup PR, but overall looks good to me. 👍

test/orbit_web/controllers/admin_controller_test.exs
skyqrose: Oh this was intentional! It's just a string comparison, rather than like a CSS selector.  I don't want to assert on what HTML attributes the button will have, so this includes the `<`, which will definitely be right before the tag name, but not the `>`, which will probably be a few attributes later.

Apologies to your balanced bracket sensibilities.

lemald: Ahhhh, I see. In that case I guess you could always match with a regular expression, otherwise I will just have to live with it I guess. 🙂

PR #103, opened 2024-08-30
Sign-in record export

lemald:
Asana Task: [📐 Operator Compliance Export for tap and type](https://app.asana.com/0/1200273269966439/1206604036091421/f)

I think there's a conversation to be had about how to best associate employees and users long-term, but for now I've opted to join them by email, since both tables have email on them and I've made sure that both have indexes on the email column. One thing to note for the future is that we _do not_ yet have a unique index on email in the employees table, but maybe we should?

One other thing I was considering is that there's at least some conceptual overlap between this and the `index` method on the `SignInController`, but the similarity in terms of code is actually only a couple / few lines of Ecto Query, so I'm not sure it's worth trying to unify those into one function.

Checklist

- Tests:
  - `(X)` Has tests
  - `( )` Doesn't need tests
- Product/Design sign off:
  - `( )` No review needed

### DIFF
--- a/js/components/home.tsx
+++ b/js/components/home.tsx
@@ -26,30 +26,31 @@ export const Home = (): ReactElement => {
         </button>
       </section>
       <section className="w-full bg-gray-100/50 py-5">
-        <div className="mx-auto max-w-lg px-2">
-          <p className="font-semibold">Export sign in records</p>
-        </div>
-        <div className="mt-5 flex justify-between gap-4 items-end mb-3 max-w-lg mx-auto px-2">
-          <label className="flex flex-grow flex-col">
-            <span className="text-sm">Service Date</span>
-            <input
-              type="date"
-              value={selectedDate}
-              onChange={(evt) => {
-                setSelectedDate(evt.target.value);
-              }}
-              className="rounded h-10 bg-gray-100"
-            />
-          </label>
-          <button
-            className="rounded border border-gray-300 hover:bg-gray-100 p-2 h-10"
-            onClick={() => {
-              console.log("TODO");
-            }}
-          >
-            Export Records
-          </button>
-        </div>
+        <form action="/sign-in-export" method="get" target="_blank">
+          <div className="mx-auto max-w-lg px-2">
+            <p className="font-semibold">Export sign in records</p>
+          </div>
+          <div className="mt-5 flex justify-between gap-4 items-end mb-3 max-w-lg mx-auto px-2">
+            <label className="flex flex-grow flex-col">
+              <span className="text-sm">Service Date</span>
+              <input
+                type="date"
+                name="date"
+                value={selectedDate}
+                onChange={(evt) => {
+                  setSelectedDate(evt.target.value);
+                }}
+                className="rounded h-10 bg-gray-100"
+              />
+            </label>
+            <button
+              className="rounded border border-gray-300 hover:bg-gray-100 p-2 h-10"
+              type="submit"
+            >
+              Export Records
+            </button>
+          </div>
+        </form>
       </section>
       <section className="max-w-lg mx-auto px-2 py-5">
         <p className="font-semibold mb-3">Today&apos;s sign ins</p>
--- a/lib/orbit_web/controllers/sign_in_export_controller.ex
+++ b/lib/orbit_web/controllers/sign_in_export_controller.ex
@@ -38,11 +38,18 @@ defmodule OrbitWeb.SignInExportController do
       )
       |> Enum.map(
         &%{
-          time: &1.signed_in_at |> DateTime.shift_zone!(@timezone) |> DateTime.to_iso8601(),
+          time:
+            &1.signed_in_at
+            |> DateTime.shift_zone!(@timezone)
+            |> Timex.format!("{YYYY}-{0M}-{0D} {0h24}:{0m}:{0s}"),
           location: "Orient Heights",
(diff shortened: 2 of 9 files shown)


js/components/home.tsx
```diff
@@ -25,11 +25,17 @@ export const Home = (): ReactElement => {
           Sign In Operator
         </button>
       </div>
-      <div className="flex justify-between gap-2 items-end mb-3">
+      <form
```
mathcolo, line 28: Wow, this is refreshing to see instead of an `onClick` attribute on the button!

lib/orbit_web/controllers/sign_in_export_controller.ex
```diff
@@ -0,0 +1,98 @@
...
+    redirect(conn, to: ~p"/sign-in-export/#{filename}")
+  end
+
+  @spec get(Plug.Conn.t(), map()) :: Plug.Conn.t()
+  def get(conn, %{"filename" => filename}) do
+    with %{"date_string" => date_string} <-
+           Regex.named_captures(~r"sign-ins-(?<date_string>.+)\.csv", filename),
+         {:ok, date} <- Date.from_iso8601(date_string) do
+      conn
+      |> put_resp_content_type("application/csv")
```
mathcolo, line 24: Per the internet, it looks like this should actually be `text/csv`?

lib/orbit_web/controllers/sign_in_export_controller.ex
```diff
@@ -0,0 +1,98 @@
...
+  end
+
+  @spec fetch_sign_ins_csv(Date.t()) :: String.t()
+  defp fetch_sign_ins_csv(service_date) do
+    {start_datetime, end_datetime} = Util.Time.service_date_boundaries(service_date)
+
+    Repo.all(
+      from operator_sign_in in OperatorSignIn,
+        where:
+          ^start_datetime < operator_sign_in.signed_in_at and
```
mathcolo, line 41: `<=`?

lib/orbit_web/controllers/sign_in_export_controller.ex
```diff
@@ -0,0 +1,98 @@
...
+            operator_sign_in.signed_in_at < ^end_datetime,
+        order_by: operator_sign_in.signed_in_at,
+        join: e in assoc(operator_sign_in, :signed_in_employee),
+        join: u in assoc(operator_sign_in, :signed_in_by_user),
+        preload: [signed_in_employee: e, signed_in_by_user: u],
+        left_join: signed_in_by_employee in Employee,
+        on: u.email == signed_in_by_employee.email,
```
mathcolo, line 48: Yeah, I think this is fine. We're sure Keycloak doesn't provide names, right?

lib/orbit_web/controllers/sign_in_export_controller.ex
```diff
@@ -0,0 +1,98 @@
...
+          operator_sign_in: operator_sign_in,
+          signed_in_by_employee: signed_in_by_employee
+        }
+    )
+    |> Enum.map(
+      &%{
+        time:
+          &1.operator_sign_in.signed_in_at
+          |> DateTime.shift_zone!(@timezone)
+          |> Timex.format!("{YYYY}-{0M}-{0D} {0h24}:{0m}:{0s}"),
```
mathcolo, line 59: Just checking, Excel is happy with this?

lib/orbit_web/controllers/sign_in_export_controller.ex
lemald: I think it might be possible to get names from SSO but I'm not sure if it's set up to pass preferred name, and in any event I think it's better to base the names displayed in the application off of the same source if possible.

lib/orbit_web/controllers/sign_in_export_controller.ex
lemald: Yes, good call - c5ff205.

lib/orbit_web/controllers/sign_in_export_controller.ex
lemald: Yeah, I'm able to load the CSV up in Excel and do date / time manipulations with this column without any additional steps to adjust the parsing.

lib/orbit_web/controllers/sign_in_export_controller.ex
```diff
@@ -0,0 +1,98 @@
...
+        location: "Orient Heights",
```
mathcolo, line 60: Should this be configurable/a param so it's easier for OL and RL?

priv/repo/migrations/20240829201608_add_employees_email_index.exs
```diff
@@ -0,0 +1,18 @@
+defmodule Orbit.Repo.Migrations.AddEmployeesEmailIndex do
+  use Ecto.Migration
+
+  def change do
+    create index(:employees, [:email])
+
+    # couple of minor stragglers from the contact -> employees rename
```
mathcolo, line 7: thank you!

lib/orbit_web/controllers/sign_in_export_controller.ex
lemald: I assume that once we have actual sign in location information it will be just another column on the `operator_sign_ins` table, so I'm not sure how I would improve this short of just using that column.

PR #104, opened 2024-08-30
refactor: look up badge number by rfid per request

skyqrose:
Asana Task: Noticed during [📐 Orbit: allow manually associating a badge serial with an operator for testing](https://app.asana.com/0/1200689710863995/1207890347662644/f), discussed in slack.

Instead of sending all badge_serials to the frontend with the employee data, this PR makes a new api endpoint to look them up one by one during sign in.

I'll need help testing this on sandbox (but I see sandbox is in use right now).

- Tests:
  - `(x)` Has tests
  - `( )` Doesn't need tests
- Product/Design sign off:
  - `(x)` No review needed

### DIFF
--- a/js/components/operatorSignIn/operatorSelection.tsx
+++ b/js/components/operatorSignIn/operatorSelection.tsx
@@ -1,7 +1,5 @@
-import { ApiResult } from "../../api";
 import { fetchEmployeeByBadgeSerial } from "../../hooks/useEmployees";
 import { useNfc } from "../../hooks/useNfc";
-import { Employee } from "../../models/employee";
 import { className } from "../../util/dom";
 import { BadgeEntry } from "./types";
 import { ReactElement, useEffect, useId, useState } from "react";
@@ -11,13 +9,11 @@ export const OperatorSelection = ({
   onOK,
   onBadgeLookupError,
   onNfcScanError,
-  employees,
 }: {
   nfcSupported: boolean;
   onOK: (badgeEntry: BadgeEntry) => void;
   onBadgeLookupError: () => void;
   onNfcScanError: () => void;
-  employees: ApiResult<Employee[]>;
 }): ReactElement => {
   const inputId = useId();
   const [badgeEntry, setBadgeEntry] = useState<BadgeEntry | null>(null);
@@ -38,7 +34,7 @@ export const OperatorSelection = ({
     } else if (nfcResult.status === "error") {
       onNfcScanError();
     }
-  }, [nfcResult, employees, onOK, onBadgeLookupError, onNfcScanError]);
+  }, [nfcResult, onOK, onBadgeLookupError, onNfcScanError]);
 
   const buttonEnabled = badgeEntry !== null;
 
--- a/js/components/operatorSignIn/operatorSelection.tsx
+++ b/js/components/operatorSignIn/operatorSelection.tsx
@@ -1,5 +1,5 @@
 import { ApiResult } from "../../api";
-import { findEmployeeByBadgeSerial } from "../../hooks/useEmployees";
+import { fetchEmployeeByBadgeSerial } from "../../hooks/useEmployees";
 import { useNfc } from "../../hooks/useNfc";
 import { Employee } from "../../models/employee";
 import { className } from "../../util/dom";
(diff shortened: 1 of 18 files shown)


lib/orbit_web/controllers/rfid_controller.ex
```diff
@@ -0,0 +1,35 @@
...
+  import Ecto.Query
+
+  alias Orbit.Repo
+
+  # TODO once we have permissions, require sign in permissions in order to use this endpoint
+
+  @spec show(Plug.Conn.t(), map()) :: Plug.Conn.t()
+  def show(conn, %{"badge_serial" => badge_serial}) do
+    badge_number =
+      Repo.one(
```
skyqrose, line 13: This will crash if a badge serial isn't unique. Which is fine for now cuz we want to fix that anyway.

js/test/api.test.ts
```diff
@@ -31,6 +31,97 @@ describe("post", () => {
...
+      json: () =>
+        new Promise((resolve) => {
+          resolve({ data: "test" });
+        }),
+    } as Response);
+
+    await expect(result).resolves.toBe("testtest");
+  });
+
+  test("rejects on parsing error", async () => {
```
lemald, line 56: **nitpick:** this test name is a big confusing because the error isn't in `parser`, maybe this would be more descriptive?
```suggestion
  test("rejects on return data type error", async () => {
```

js/components/operatorSignIn/operatorSelection.tsx
```diff
@@ -25,17 +25,16 @@ export const OperatorSelection = ({
...
-      )?.badge;
-
-      if (badge) {
-        onOK({ number: badge, method: "nfc" });
-      } else {
-        onBadgeLookupError();
-      }
+    if (nfcResult.status === "success") {
+      const badgeSerial = nfcResult.data;
+      fetchEmployeeByBadgeSerial(badgeSerial).then(
```
lemald, line 30: **suggestion:** I do want to flag that there's a potential race condition here where the `useEffect` dependencies change before this promise resolves. I think that because of the way the `useNfc` hook works in practice (it "locks in" the result of a successful scan) it's unlikely that it will cause an actual problem in practice but I haven't yet 100% convinced myself of that.

test/support/factory.ex
```diff
@@ -16,7 +16,7 @@ defmodule Orbit.Factory do
       middle_initial: "A",
       email: sequence(:employee_email, &"fake#{&1}@test.com"),
       badge_number: sequence("employee_badge"),
-      badge_serials: [build(:badge_serial)],
```
lemald, line 19: **question:** what's the purpose of removing this from the factory? We should still have the badge serials available on the back-end, it's just that the controller won't preload them in its query and won't include a `badge_serials` key in its response.

js/components/operatorSignIn/operatorSelection.tsx
skyqrose: Yeah, I'd been thinking about that too. I was leaning towards it being unlikely enough to pop up that I just didn't bother to put in the effort to fix it. Maybe I should leave a comment nearby acknowledging that the race condition exists, even if I don't fix it now?

lemald: A comment sounds good! I think that the way that the `useNfc` hook works is the key part of the rationale here to mention.

test/support/factory.ex
skyqrose: No serials seemed like a simpler default value. Usually if you care that this field exists, you should set it in the test, so that changing the factory values doesn't break tests. There was only one test that depended on this and I ended up editing it anyway.

lemald (approved):
There's still my [nitpick comment about test naming](https://github.com/mbta/orbit/pull/104#discussion_r1739306397) but aside from that this looks good.

js/test/api.test.ts
skyqrose: I think the step that's failing could be called parsing, and zod even calls this `zod.parse()`. I think it's actually `parser` that's misnamed, since that's about transforming typed data into other typed data, rather than turning unknown data into typed data. (The name is left over from before we used zod in Glides, and that step actually did some parsing.)

skyqrose:
I think I want to wait until Tuesday to merge this so we have a chance to do an end-to-end test of this on sandbox. (I'm not set up to test it locally.)

js/test/api.test.ts
lemald: I guess I'd say that really neither should be called parsing strictly speaking because they're not about taking a string and tokenizing it and turning it into an abstract syntax tree, but if Zod uses the terminology then _I guess_ that's okay and it's not a hill I'll choose to die on. 😝

skyqrose:
Tested this manually with @mathcolo and it works.

PR #105, opened 2024-08-30
refactor: frontend employee name fields

skyqrose:
Asana Task: None. Noticed while doing #104

Stacked on top of #104.

Things this changes:
- first and last name can be nil on the backend, but the frontend assumed they'd always be there. This PR makes them nullable and handles if they're missing.
- The non-preferred first name isn't used on the frontend, so might as well not send it at all, for better privacy and so the frontend doesn't have to juggle the two of them.
- Except for the one spot we accidentally did use first_name instead of preferred_first. Now it'll use the preferred name if available.
- DRYs one place where `lookupDisplayName` was reimplemented.

- Tests:
  - `(x)` Has tests
  - `( )` Doesn't need tests
- Product/Design sign off:
  - `(x)` No review needed

### DIFF
--- a/js/components/operatorSignIn/operatorSignInModal.tsx
+++ b/js/components/operatorSignIn/operatorSignInModal.tsx
@@ -80,10 +80,10 @@ export const OperatorSignInModal = ({
   }, [complete, close]);
 
   const employee =
-    employees.status === "ok" &&
-    badge !== null &&
-    findEmployeeByBadge(employees.result, badge.number);
-  const name = employee ? employee.first_name : `Operator ${badge?.number}`;
+    employees.status === "ok" && badge !== null ?
+      findEmployeeByBadge(employees.result, badge.number)
+    : null;
+  const name: string = employee?.first_name ?? `Operator ${badge?.number}`;
 
   return (
     <Modal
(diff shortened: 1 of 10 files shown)


js/components/operatorSignIn/operatorSignInModal.tsx
```diff
@@ -80,10 +80,10 @@ export const OperatorSignInModal = ({
...
 
   const employee =
-    employees.status === "ok" &&
-    badge !== null &&
-    findEmployeeByBadge(employees.result, badge.number);
-  const name = employee ? employee.first_name : `Operator ${badge?.number}`;
+    employees.status === "ok" && badge !== null ?
+      findEmployeeByBadge(employees.result, badge.number)
+    : null;
+  const name: string = employee?.first_name ?? `Operator ${badge?.number}`;
```
skyqrose, line 86: I was wondering if we should DRY this? It's a little different than `lookupDisplayName` because it only uses the first name.

lemald: Eh I think it's fine given that, as you point out, it's a bit different.

lemald (approved):

PR #111, opened 2024-09-03
unique badge serials

skyqrose:
Asana Task: [📐 Orbit: allow manually associating a badge serial with an operator for testing](https://app.asana.com/0/1200689710863995/1207890347662644/f)

Migration to make badge serials unique, and updates to the two ways to create badge serials (/admin/rfid and the csv import) to enforce that.

- Tests:
  - `(x)` Has tests
  - `( )` Doesn't need tests
- Product/Design sign off:
  - `( )` No review needed

### DIFF
--- a/lib/orbit/import/rfid.ex
+++ b/lib/orbit/import/rfid.ex
@@ -1,5 +1,7 @@
 defmodule Orbit.Import.Rfid do
   require Logger
+  import Ecto.Query
+  alias Orbit.BadgeSerial
   alias Orbit.Employee
   alias Orbit.Repo
   alias Orbit.S3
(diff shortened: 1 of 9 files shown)


lib/orbit/import/rfid.ex
```diff
@@ -20,39 +22,56 @@ defmodule Orbit.Import.Rfid do
...
+
+    if num_serials_with_duplicates > 0 do
+      Logger.warning("rfid_import badge_serials_with_duplicates=#{num_serials_with_duplicates}")
+    end
+
+    serials_by_badge =
+      Enum.group_by(rows, & &1[@badge_number_field], & &1[@badge_serial_field])
+
+    {:ok, imported_count} =
+      Repo.transaction(fn ->
```
lemald, line 44: **musing (non-blocking):** I'm sad that this means we can't just easily leverage built-in Ecto functionality and the `on_replace` argument to the `has_many` association, but given the nature of the constraint across badge serials for different employees I don't think there's really a better way to do it. In any event I think this transaction should do the trick of making the changes atomic, and I'm not too worried about race conditions because there isn't really other functionality writing this table concurrently (especially because the employee info and RFID imports are offset by half an hour).

lemald (approved):

PR #113, opened 2024-09-03
fix: fixes for mobile rendering

mathcolo:
Asana Task: [🛠 Orbit Polish](https://app.asana.com/0/1206105669438487/1208045647859196/f)

Some more styling to help with mobile.

**Table display:**
This introduces a soft hyphen around @ signs, if present, so a full email address breaks there. I tried a lot of other solutions, including:
- Splitting `@` parts into different elements; works, but it's complex/annoying
- Zero width space: annoying, because it can land on the pasteboard if copied
- This solution, but setting `hyphenate-character` to '' via inline CSS; bundler removed this I think?

image

Checklist

- Tests:
  - `( )` Has tests
  - `(x)` Doesn't need tests
- Product/Design sign off:
  - `( )` No review needed

### DIFF
--- a/js/components/operatorSignIn/operatorSelection.tsx
+++ b/js/components/operatorSignIn/operatorSelection.tsx
@@ -48,7 +48,7 @@ export const OperatorSelection = ({
       <Or />
       <label htmlFor={inputId}>Search for an Operator</label>
       <input
-        className="h-[61px]"
+        className="h-[40px]"
         id={inputId}
         inputMode="numeric"
         onChange={(evt) => {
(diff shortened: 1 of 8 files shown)


lemald:
> This solution, but setting hyphenate-character to '' via inline CSS; bundler removed this I think?

Did you try to put this in a `style` attribute, or did you try using Tailwind's support for [arbitrary properties](https://tailwindcss.com/docs/adding-custom-styles#arbitrary-properties)?

mathcolo:
@lemald It was inline `style`. I'll dig into Tailwind's option 👍

js/components/operatorSignIn/operatorSelection.tsx
```diff
@@ -48,7 +48,7 @@ export const OperatorSelection = ({
       <Or />
       <label htmlFor={inputId}>Search for an Operator</label>
       <input
-        className="h-[61px]"
+        className="h-[40px]"
```
lemald, line 51: **suggestion:** since 40px is in the [default spacing scale](https://tailwindcss.com/docs/customizing-spacing#default-spacing-scale) could you just do this instead?
```suggestion
        className="h-10"
```

mathcolo: ah yes!

lemald (approved):
👍

PR #117, opened 2024-09-05
feat: navigation menu + wordmark

mathcolo:
Asana Task: [📐 Implement navigation menu + wordmark](https://app.asana.com/0/1206105669438487/1207966100310392/f)

image

Checklist

- Tests:
  - `(x)` Has tests
  - `( )` Doesn't need tests
- Product/Design sign off:
  - `( )` No review needed


js/components/header.tsx
```diff
@@ -1,10 +1,32 @@
...
+    currentLocation.pathname === paths.menu ||
+    currentLocation.pathname === paths.help;
+
   return (
     <div className="w-full bg-gray-200 p-2 flex justify-between">
-      <img src="/images/logo.svg" alt="MBTA" className="w-8" />
-      <a href="/logout" className="underline p-1">
-        Log out
-      </a>
+      <Link to={"/"}>
```
lemald, line 13: **suggestion:** is there some reason this can't just be
```suggestion
      <Link to={paths.root}>
```
now?

js/components/menus.tsx
```diff
@@ -0,0 +1,111 @@
...
+const MenuLink = ({
+  to,
+  icon,
+  iconExtraClassName,
+  reload,
+  newTab,
+  children,
+}: {
+  to: string;
+  icon?: string;
```
lemald, line 16: **suggestion:** rather than an arbitrary string, could you define a custom `IconPath` type (or something like that) that only allows specifying valid icons?

lemald: Even if it's just something simple manually defined like `type IconName = "foo" | "bar" | ...`.

lib/orbit/authentication/user.ex
```diff
@@ -38,4 +42,17 @@ defmodule Orbit.Authentication.User do
...
+    List.first(
+      Repo.all(
+        from(e in Employee,
+          where: e.email == ^struct.email,
+          select: %{first_name: e.first_name, preferred_first: e.preferred_first}
+        )
+      ),
+      nil
+    )
+  end
```
lemald, line 57: **suggestion:** if we always plan on applying logic to the result to use the preferred name if present, should we just do that here? And if so, maybe call this something a bit more specific like `get_display_name_from_employee`?

lemald: Also I think it would be good to somehow prevent the case of duplicate emails going forward, maybe in a separate ticket / PR.

test/orbit/authentication/user_test.exs
```diff
@@ -43,4 +45,50 @@ defmodule Orbit.Authentication.UserTest do
...
+        last_name: "Read",
+        email: "arthur@mbta.com",
+        badge_number: "123456789"
+      })
+
+      user =
+        Repo.insert!(%User{
+          email: "arthur@mbta.com",
+          permissions: [:operator_sign_in]
+        })
```
lemald, line 63: **suggestion:** these could both be accomplished a bit more idiomatically with the `insert!` macro from ExMachina and our existing factories.

js/components/header.tsx
mathcolo: nope! a423717

js/components/menus.tsx
mathcolo: 👍 60571af

lib/orbit/authentication/user.ex
mathcolo: e7b8ea0

I agree re: duplicate 👍 I'll file a task

test/orbit/authentication/user_test.exs
mathcolo: 👍 7f75eed

lemald (approved):

PR #123, opened 2024-09-10
feat: auto-update signins list on home page

mathcolo:
Asana Task: [🐞 Orbit 'Today's sign ins' table doesn't auto-update with new sign-in on same device](https://app.asana.com/0/1206105669438487/1208214593975954/f)

Change the home component so the "Today's sign-ins" list reloads after we know there's a new sign-in on the server.

Checklist

- Tests:
  - `(x)` Has tests
  - `( )` Doesn't need tests
- Product/Design sign off:
  - `( )` No review needed


js/components/home.tsx
```diff
@@ -54,11 +59,14 @@ export const Home = (): ReactElement => {
       </section>
       <section className="max-w-lg mx-auto px-2 py-5">
         <p className="font-semibold mb-3">Today&apos;s sign ins</p>
-        <List line="blue" />
+        <List key={signInCount} line="blue" />
```
mathcolo, line 62: I spent a couple of hours trying to test _specifically_ that the state increment causes the list to reload. It's theoretically possible, but I got into a mess of mocking the modal and trying to trigger `onComplete` with a fake button. I'm submitting this PR for now and I can re-visit if we think it's absolutely necessary.

js/components/operatorSignIn/operatorSignInModal.tsx
```diff
@@ -138,6 +146,7 @@ const OperatorSignInModalContent = ({
           loading={loading}
           onComplete={() => {
             submit(badge, setComplete, setLoading);
+            onComplete();
```
lemald, line 149: **question:** as I understand it, `submit` returns a promise but we're not actually waiting for that to resolve before calling `onComplete`, right? I worry that that might trigger a race condition where we might call `onComplete` and refresh the page but do so just quickly enough that it doesn't pick up the just-submitted data.

js/components/home.tsx
```diff
@@ -54,11 +59,14 @@ export const Home = (): ReactElement => {
...
       </section>
 
       <OperatorSignInModal
         show={modalOpen}
+        onComplete={() => {
+          setSignInCount(signInCount + 1);
```
lemald, line 68: **suggestion:** this is a case where you can pass a callback to a setter instead of a new value:
```suggestion
          setSignInCount((count) => count + 1);
```

I don't think it makes a huge difference here but I think it's vaguely more idiomatic and can be nice in certain situations (for instance, inside of a `useEffect` it can avoid having to add the old value as a depndency).

js/components/operatorSignIn/operatorSignInModal.tsx
lemald: One option would be to just make this callback another argument to the `submit` function which could then invoke it at the appropriate part of the lifecycle of the promise once the HTTP request has gone through.

js/components/home.tsx
mathcolo: nice! fixed fd3dcde

js/components/operatorSignIn/operatorSignInModal.tsx
mathcolo: Ooh, good call out. I blanked that's in the promised land. fd3dcde

js/components/home.tsx
lemald: I'd be curious to learn / chat more about the approach you were trying and what issues you were running into.

mathcolo: (We talked on Slack)
This is a lot better! bb2b335

js/test/components/home.test.tsx
```diff
@@ -16,4 +42,19 @@ describe("home", () => {
...
+    const view = render(
+      <MemoryRouter>
+        <Home />
+      </MemoryRouter>,
+    );
+    await userEvent.click(
+      view.getByRole("button", { name: "Sign In Operator" }),
+    );
+
+    (fetch as jest.MockedFunction<typeof fetch>).mockClear();
```
lemald, line 56: **question:** is it possible to use `jest.mocked(...)` here instead of having to use `as`?

js/test/components/home.test.tsx
```diff
@@ -16,4 +42,19 @@ describe("home", () => {
...
+    await userEvent.click(view.getByRole("button", { name: "Complete" }));
+    expect(fetch).toHaveBeenCalledWith("/api/signin?line=blue");
```
lemald, line 58: **question:** just to clarify, is the point of the `mockClear` above so that we can be sure that this `expect` is catching the reload and not the fetch that happens on the initial render?

mathcolo: Yes 👍 I'll note that in a comment

js/test/components/home.test.tsx
mathcolo: I KNEW there was something I forgot here. 56c3aef

js/test/components/home.test.tsx
mathcolo: 56c3aef

lemald (approved):

PR #128, opened 2024-09-11
feat: Install FullStory

mathcolo:
Asana Task: [🛠 Install Fullstory](https://app.asana.com/0/1206105669438487/1207316362183269/f)

- Adds FullStory JS packages
- Rename `sentry.js` to `telemetry.js`; adds FullStory config to it
- Adds FullStory domain to CSP header
- Ties FULLSTORY_ORG_ID through

Checklist

- [x] - ⚠️ Requires https://github.com/mbta/terraform_modules/pull/43
- [x] - Add domains to fullstory.com config
- Tests:
  - `( )` Has tests
  - `(x)` Doesn't need tests **(just doesn't seem valuable here)**
- Product/Design sign off:
  - `( )` No review needed


lib/orbit_web/components/layouts/root.html.heex
```diff
@@ -23,6 +23,9 @@
     <%= if @conn.assigns[:name] do %>
       <meta name="userName" content={@name} />
     <% end %>
+    <%= if full_story_org_id = Application.get_env(:orbit, :full_story_org_id) do %>
```
lemald, line 26: **suggestion / nitpick:** I think it would be a bit more idiomatic, as well as consistent with the other `<meta>` tags, if we put the FullStory org ID in the `conn` assigned rather than directly calling `get_env` here.

mathcolo: agreed- f43f3ff

lemald (approved):

PR #137, opened 2024-09-16
feat: osi list shows officials' name with email as fallback

mathcolo:
Asana Task: [📐 Orbit: In-app view of completed sign-ins does not include official full name](https://app.asana.com/0/1206105669438487/1208171871591943/f)

Display officials' names in the list, with email address as fallback.

Checklist

- Tests:
  - `(x)` Has tests
  - `( )` Doesn't need tests
- Product/Design sign off:
  - `( )` No review needed

### DIFF
--- a/lib/orbit_web/controllers/sign_in_controller.ex
+++ b/lib/orbit_web/controllers/sign_in_controller.ex
@@ -31,16 +31,28 @@ defmodule OrbitWeb.SignInController do
                 ^start_datetime <= si.signed_in_at and
                   si.signed_in_at < ^end_datetime and
                   si.rail_line == ^rail_line,
-              preload: [:signed_in_employee, :signed_in_by_user],
-              order_by: [desc: :signed_in_at]
+              join: u in assoc(si, :signed_in_by_user),
+              preload: [:signed_in_employee, signed_in_by_user: u],
+              left_join: signed_in_by_employee in Employee,
+              on: u.email == signed_in_by_employee.email,
+              order_by: [desc: :signed_in_at],
+              select: %{
+                signin: si,
+                signed_in_by_employee: signed_in_by_employee
+              }
             )
           )
-          |> Enum.map(fn si ->
+          |> Enum.map(fn i ->
             %{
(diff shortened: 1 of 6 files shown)


lib/orbit_web/controllers/sign_in_controller.ex
```diff
@@ -31,16 +31,28 @@ defmodule OrbitWeb.SignInController do
                 ^start_datetime <= si.signed_in_at and
                   si.signed_in_at < ^end_datetime and
                   si.rail_line == ^rail_line,
-              preload: [:signed_in_employee, :signed_in_by_user],
-              order_by: [desc: :signed_in_at]
+              join: u in assoc(si, :signed_in_by_user),
+              preload: [:signed_in_employee, signed_in_by_user: u],
```
lemald, line 35: **suggestion (non-blocking):** If you're already in here turning the preload for `signed_in_by_user` from the default (which uses a separate query - see [here](https://hexdocs.pm/ecto/Ecto.Query.html#preload/3)) to using an explicit join in the same query, it might be worth doing the same for `signed_in_employee`, thus turning this whole operation into a single query.

lemald (approved):
One suggestion for a further improvement you could make but overall looks good to me. 👍

lib/orbit_web/controllers/sign_in_controller.ex
mathcolo: ah! thanks, I see you did that in https://github.com/mbta/orbit/pull/103. why is that not the default preload behavior, though? I guess because Ecto wants joins to remain explicit?

mathcolo: f8ad95b

lemald: > why is that not the default preload behavior, though?

I'm not sure - it's also possible that Ecto isn't able to generate the join statements fully programmatically or something? In any event it's just an interesting quirk to be aware of.

lemald (approved):

PR #141, opened 2024-09-20
fix(OperatorSelection): ignore leading zeroes

mathcolo:
Asana Task: [Allow leading zero in Orbit type workaround](https://app.asana.com/0/1206105669438487/1208362994470250/f)

Ignore leading zero in operator selection & attestation, as we cannot tell what personnel are used to.

Checklist

- Tests:
  - `(x)` Has tests
  - `( )` Doesn't need tests
- Product/Design sign off:
  - `( )` No review needed

### DIFF
--- a/js/components/operatorSignIn/attestation.tsx
+++ b/js/components/operatorSignIn/attestation.tsx
@@ -158,6 +158,7 @@ export const SignaturePrompt = ({
           className="fs-mask w-full"
           inputMode="numeric"
           defaultValue={defaultValue}
+          // Do not set `value`- we are transforming below!
           onChange={(evt) => {
             onChange(removeLeadingZero(evt.target.value));
           }}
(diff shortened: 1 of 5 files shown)


js/components/operatorSignIn/attestation.tsx
```diff
@@ -158,7 +159,7 @@ export const SignaturePrompt = ({
           inputMode="numeric"
           defaultValue={defaultValue}
           onChange={(evt) => {
-            onChange(evt.target.value);
+            onChange(removeLeadingZero(evt.target.value));
```
skyqrose, line 162: Hmm, doing it this way means the react state is different from what's in the input. I think it'd work, but it seems like it has potential to end up with weird bugs if we change anything, like the input not visually responding if you type a 0.

An alternative would be to include the text as typed, including leading 0s, in the react state. And then remove the leading zeroes wherever we use that variable.

mathcolo: Point taken 👍 will fix

mathcolo: discussed in slack- adding comments instead of code change to prevent `removeLeadingZero` from having to make it to the modal as well.

skyqrose (approved):

PR #148, opened 2024-10-03
DO NOT MERGE: import all personnel, not just area 114

mathcolo:
Asana Task: [🛠 (Orbit Sandbox) Allow anyone with a valid badge to be signed in fit for duty](https://app.asana.com/0/1206105669438487/1208389751505959/f)

Description goes here.
Include screenshots if relevant.

Checklist

- Tests:
  - `( )` Has tests
  - `( )` Doesn't need tests
- Product/Design sign off:
  - `( )` No review needed

### DIFF
--- a/config/prod.exs
+++ b/config/prod.exs
@@ -18,9 +18,9 @@ config :orbit, Oban,
      timezone: "America/New_York",
      crontab: [
        # the top of every hour, any day, any month, any day of the week
-       {"0 * * * *", Orbit.Import.PersonnelWorker},
+       {"*/5 * * * *", Orbit.Import.PersonnelWorker},
        # the bottom of every hour, any day, any month, any day of the week
-       {"30 * * * *", Orbit.Import.RfidWorker}
+       {"*/5 * * * *", Orbit.Import.RfidWorker}
      ]}
   ]
 
--- a/lib/orbit/import/personnel.ex
+++ b/lib/orbit/import/personnel.ex
@@ -43,12 +43,16 @@ defmodule Orbit.Import.Personnel do
           area: String.to_integer(&1[@area_value_field])
         }
       )
-      |> Enum.map(
-        &Repo.insert(&1,
-          on_conflict: {:replace_all_except, [:id]},
-          conflict_target: :badge_number
-        )
-      )
+      |> Enum.map(fn item ->
+        try do
+          Repo.insert(item,
+            on_conflict: {:replace_all_except, [:id]},
+            conflict_target: :badge_number
+          )
+        rescue
+          Postgrex.Error -> nil
+        end
+      end)
 
     Logger.info("personnel_import count=#{length(result)}")
   end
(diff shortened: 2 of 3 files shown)


config/prod.exs
```diff
@@ -18,9 +18,9 @@ config :orbit, Oban,
      timezone: "America/New_York",
      crontab: [
        # the top of every hour, any day, any month, any day of the week
-       {"0 * * * *", Orbit.Import.PersonnelWorker},
+       {"*/5 * * * *", Orbit.Import.PersonnelWorker},
        # the bottom of every hour, any day, any month, any day of the week
```
skyqrose, line 22: these comment descriptions are no longer accurate (that's fine, you don't need to update comments for temporary stuff like this)

Was this change just so you don't have to wait an hour to test it?

lib/orbit/import/personnel.ex
```diff
@@ -43,12 +43,16 @@ defmodule Orbit.Import.Personnel do
...
-        )
-      )
+      |> Enum.map(fn item ->
+        try do
+          Repo.insert(item,
+            on_conflict: {:replace_all_except, [:id]},
+            conflict_target: :badge_number
+          )
+        rescue
+          Postgrex.Error -> nil
```
skyqrose, line 53: Do you expect this to happen?

lib/orbit/import/rfid.ex
```diff
@@ -41,53 +41,61 @@ defmodule Orbit.Import.Rfid do
...
-    |> String.splitter("\n", trim: true)
-    |> CSV.decode!(headers: true, field_transform: &String.trim/1)
-    |> Rfid.import_rows()
+    try do
+      Rfid.download()
+      |> String.splitter("\n", trim: true)
+      |> CSV.decode!(headers: true, field_transform: &String.trim/1)
+      |> Rfid.import_rows()
+    rescue
+      error -> Logger.error(inspect(error))
```
skyqrose, line 97: Why do you need to squash errors here?

config/prod.exs
mathcolo: Yep!

lib/orbit/import/personnel.ex
mathcolo: Yes; there's one preferred name that contains an invalid byte, so I'm skipping it for now

lib/orbit/import/rfid.ex
mathcolo: We don't have Oban telemetry set up yet (https://app.asana.com/0/616151179860796/1207721363846631/f) so I wasn't seeing anything in Splunk. Needed to do this to quickly see I was exceeding the 15s transaction timeout

lib/orbit/import/personnel.ex
skyqrose: ok! "catch all and ignore" scares me in general, but it's fine for the proof of concept.

Should we handle this case for real in Real Orbit? What's the name and the invalid byte issue? (DM me the name, don't put it here.) If it's a text encoding issue for a real name, we should see if we can get Orbit to handle it. If it's something silly like HR/CMS storing invalid UTF-8, then I guess we could filter it out before inserting.

PR #163, opened 2024-10-23
feat: Radio backend + reorder CSV

mathcolo:
Asana Task: [📐 Orbit radio export implementation](https://app.asana.com/0/1206105669438487/1208470646928351/f)

- Add `radio_number` column to database
- CSV: Add `Radio`
- CSV: Reorder

Checklist

- Tests:
  - `(x)` Has tests
  - `( )` Doesn't need tests
  - `( )` Tests deferred (with justification)
- Product/Design sign off:
  - `(x)` Okayed the plan for the feature (e.g. the design files, or the Asana task)
  - `( )` Reviewed the feature as implemented (e.g. on dev-green, or saw screenshots)
  - `( )` No review needed

PR #164, opened 2024-10-23
feat: Radio frontend

mathcolo:
Asana Task: [📐 Orbit radio UI implementation](https://app.asana.com/0/1206105669438487/1208470630634721/f)

Still need to do some testing on mobile (as of this writing I don't have my phone accessible), but otherwise this is radio added to the frontend list display + form pages in a #163-not-ready-yet compatible way. That PR should be merged first.

image

Checklist

- Tests:
  - `(x)` Has tests
  - `( )` Doesn't need tests
- Product/Design sign off:
  - `( )` No review needed

### DIFF
--- a/js/components/operatorSignIn/attestation.tsx
+++ b/js/components/operatorSignIn/attestation.tsx
@@ -51,31 +51,33 @@ export const Attestation = ({
     <div className="text-sm">
       Step 2 of 2
       <SignInText />
-      <InputBox
-        title={"Operator Badge Number"}
-        defaultValue={defaultValue}
-        onChange={setEnteredBadge}
-      />
-      <SignatureHint badge={badge} signatureText={enteredBadge} />
-      <InputBox
-        title={"Radio Number"}
-        defaultValue={radio}
-        onChange={setRadio}
-      />
-      <p className="my-3">
-        By pressing the button below I, <b className="fs-mask">{name}</b>,
-        confirm the above is true.
-      </p>
-      <button
-        className={className([
-          "block w-full md:max-w-64 mx-auto h-10 px-5 bg-gray-500 text-gray-200 rounded-md",
-          (!valid || loading) && "opacity-50",
-        ])}
-        onClick={onComplete}
-        disabled={!valid}
-      >
-        Complete Fit for Duty Check
-      </button>
+      <form>
+        <InputBox
+          title={"Operator Badge Number"}
+          defaultValue={defaultValue}
+          onChange={setEnteredBadge}
+        />
+        <SignatureHint badge={badge} signatureText={enteredBadge} />
+        <InputBox
+          title={"Radio Number"}
+          defaultValue={radio}
+          onChange={setRadio}
+        />
+        <p className="my-3">
+          By pressing the button below I, <b className="fs-mask">{name}</b>,
+          confirm the above is true.
+        </p>
+        <button
+          className={className([
+            "block w-full md:max-w-64 mx-auto h-10 px-5 bg-gray-500 text-gray-200 rounded-md",
+            (!valid || loading) && "opacity-50",
+          ])}
+          onClick={onComplete}
+          disabled={!valid}
+        >
+          Complete Fit for Duty Check
+        </button>
+      </form>
     </div>
   );
 };
--- a/js/components/operatorSignIn/attestation.tsx
+++ b/js/components/operatorSignIn/attestation.tsx
@@ -5,7 +5,7 @@ import { Employee } from "../../models/employee";
 import { className } from "../../util/dom";
 import { removeLeadingZero } from "../../util/string";
 import { useSignInText } from "./text";
-import { ChangeEvent, ReactElement, useEffect, useState } from "react";
+import { ReactElement, useEffect, useState } from "react";
 
 export const Attestation = ({
   badge,
@@ -56,16 +56,16 @@ export const Attestation = ({
         <InputBox
           title={"Operator Badge Number"}
           defaultValue={defaultValue}
-          onChange={(evt) => {
-            setEnteredBadge(removeLeadingZero(evt.target.value));
+          onChange={(value) => {
+            setEnteredBadge(removeLeadingZero(value));
           }}
         />
         <SignatureHint badge={badge} signatureText={enteredBadge} />
         <InputBox
           title={"Radio Number"}
           defaultValue={""}
-          onChange={(evt) => {
-            setEnteredRadio(evt.target.value);
+          onChange={(value) => {
+            setEnteredRadio(value);
           }}
         />
         <p className="my-3">
@@ -161,12 +161,12 @@ const SignatureHint = ({
   );
 };
 
-export const InputBox = ({
+const InputBox = ({
   onChange,
   defaultValue,
   title,
 }: {
-  onChange: (evt: ChangeEvent<HTMLInputElement>) => void;
+  onChange: (value: string) => void;
   defaultValue: string;
   title: string;
 }): ReactElement => {
@@ -183,7 +183,9 @@ export const InputBox = ({
           inputMode="numeric"
           defaultValue={defaultValue}
           // Do not set `value`- we are transforming below!
-          onChange={onChange}
+          onChange={(evt) => {
+            onChange(evt.target.value);
+          }}
           required
         />
       </label>
--- a/js/components/operatorSignIn/attestation.tsx
+++ b/js/components/operatorSignIn/attestation.tsx
@@ -51,7 +51,11 @@ export const Attestation = ({
     <div className="text-sm">
       Step 2 of 2
       <SignInText />
-      <form>
+      <form
+        onSubmit={(e) => {
+          e.preventDefault();
+        }}
+      >
         <InputBox
           title={"Operator Badge Number"}
           defaultValue={defaultValue}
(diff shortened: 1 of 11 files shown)


js/components/operatorSignIn/list.tsx
```diff
@@ -39,6 +40,10 @@ export const List = ({ line }: { line: HeavyRailLine }): ReactElement => {
             <td className="fs-mask border-y md:border-x p-1 break-all">
               {si.signed_in_employee}
             </td>
+            <td className="fs-mask border-y md:border-x p-1 break-all">
+              {/* {si.radio_number} */}
+              2222
```
mathcolo, line 45: This is why tests are failing.

js/models/signin.ts
```diff
@@ -2,6 +2,8 @@ import { z } from "zod";
 
 export const SignIn = z.object({
   rail_line: z.enum(["blue", "orange", "red"]),
+  // Optional until the backend change is merged
```
skyqrose, line 5: It's not optional from a user-perspective, but is it actually always going to be there from a data perspective? What if we use this model to view an old sign in?

js/components/operatorSignIn/list.tsx
skyqrose: Related to this field being nullable from a data perspective, maybe this should gracefully display an empty/missing state if the radio number is missing?

js/components/operatorSignIn/list.tsx
```diff
@@ -39,6 +40,10 @@ export const List = ({ line }: { line: HeavyRailLine }): ReactElement => {
             <td className="fs-mask border-y md:border-x p-1 break-all">
               {si.signed_in_employee}
             </td>
+            <td className="fs-mask border-y md:border-x p-1 break-all">
```
skyqrose, line 43: Why `fs-mask`? This doesn't seem obviously sensitive like a name.

js/components/operatorSignIn/attestation.tsx
```diff
@@ -47,22 +51,37 @@ export const Attestation = ({
...
-      </button>
+        <InputBox
+          title={"Operator Badge Number"}
+          defaultValue={defaultValue}
+          onChange={setEnteredBadge}
+        />
+        <SignatureHint badge={badge} signatureText={enteredBadge} />
+        <InputBox
+          title={"Radio Number"}
+          defaultValue={radio}
```
skyqrose, line 67: Shouldn't the default be `""`?

js/components/operatorSignIn/attestation.tsx
```diff
@@ -47,22 +51,37 @@ export const Attestation = ({
...
       >
-        Complete Fit for Duty Check
-      </button>
+        <InputBox
+          title={"Operator Badge Number"}
+          defaultValue={defaultValue}
+          onChange={setEnteredBadge}
+        />
+        <SignatureHint badge={badge} signatureText={enteredBadge} />
+        <InputBox
```
skyqrose, line 65: Do you still want the `removeLeadingZero` behavior in the radio input?

js/test/components/operatorSignIn/attestation.test.tsx
```diff
@@ -97,17 +136,32 @@ describe("Attestation", () => {
   });
 
   test("valid attestation", async () => {
+    // 'Radio' input state is held a component level up to keep
+    //   retry easy. Therefore we're pushing "22" in as a prop,
+    //   but also entering it in the test to make sure setRadio is cadlled.
```
skyqrose, line 141: typo `cadlled`

js/test/components/operatorSignIn/attestation.test.tsx
```diff
@@ -97,17 +136,32 @@ describe("Attestation", () => {
   });
 
   test("valid attestation", async () => {
+    // 'Radio' input state is held a component level up to keep
+    //   retry easy. Therefore we're pushing "22" in as a prop,
```
skyqrose, line 140: I wonder if it's possible to set this up more realistically? Something like
```tsx
const Wrapper = () => {
  [radio, setRadio] = useState("");
  return <Attestation radio={radio} setRadio={setRadio} />
}
render(<Wrapper />);
```

js/components/operatorSignIn/attestation.tsx
mathcolo: This mattered less when radio number was a string; fixed now. thanks! 0f77073

js/test/components/operatorSignIn/attestation.test.tsx
mathcolo: I changed this so Attestation submits the radio number via the `onComplete`, which the modal can then use to set its state as well as submit the sign-in. Seems better!

js/components/operatorSignIn/list.tsx
mathcolo: doing so now 👍

js/components/operatorSignIn/attestation.tsx
```diff
@@ -139,17 +161,19 @@ const SignatureHint = ({
...
 };
 
-export const SignaturePrompt = ({
+export const InputBox = ({
   onChange,
   defaultValue,
+  title,
 }: {
-  onChange: (value: string) => void;
+  onChange: (evt: ChangeEvent<HTMLInputElement>) => void;
```
skyqrose, line 169: optional refactoring suggestion:
`InputBox` isn't used outside of this file, so doesn't need the `export`.
`onChange` might be cleaner if it passed the `evt.target.value` as a `string`, instead of passing the whole event and making the caller get the value. That way the event details stay closer to the HTML input that generated them.

js/components/operatorSignIn/attestation.tsx
```diff
@@ -47,22 +48,43 @@ export const Attestation = ({
...
+          By pressing the button below I, <b className="fs-mask">{name}</b>,
+          confirm the above is true.
+        </p>
+        <button
+          className={className([
+            "block w-full md:max-w-64 mx-auto h-10 px-5 bg-gray-500 text-gray-200 rounded-md",
+            (!valid || loading) && "opacity-50",
+          ])}
+          onClick={() => {
+            onComplete(enteredRadio);
```
skyqrose, line 81: optional refactoring idea: I remember there was a thing about using form `onSubmit` instead of the button `onClick` for something else recently. Would it be better to make this button a submit button and do this call in the form's `onSubmit` handler instead too?

js/components/operatorSignIn/attestation.tsx
```diff
@@ -16,14 +16,15 @@ export const Attestation = ({
 }: {
   badge: string;
   employees: ApiResult<Employee[]>;
-  onComplete: () => void;
+  onComplete: (radio: string) => void;
```
skyqrose, line 19: praise: passing the radio on complete instead of trying to keep the state updated in the parent is way better, good call on that.

js/components/operatorSignIn/operatorSignInModal.tsx
```diff
@@ -94,6 +96,7 @@ const OperatorSignInModalContent = ({
   close: () => void;
 }): ReactElement => {
   const [badge, setBadge] = useState<BadgeEntry | null>(null);
+  const [radio, setRadio] = useState<string>("");
```
skyqrose, line 99: thought: I don't love how this component still requires a radio state just for tryAgain. But I don't see a different way to do that, so 👍

js/test/components/operatorSignIn/attestation.test.tsx
```diff
@@ -77,7 +80,29 @@ describe("Attestation", () => {
...
-      expect(view.getByRole("textbox")).toHaveValue("123");
+      const input = view.getByLabelText(/Operator Badge Number/i, {
+        selector: "input",
+      });
+      expect(input).toHaveValue("123");
+    });
+  });
+
+  describe("radio text box", () => {
+    test("it's there", () => {
```
skyqrose, line 91: The presence of the input is already tested by the other tests that also test filling it out. So I don't think this test adds anything and it could be removed.

js/test/components/operatorSignIn/attestation.test.tsx
```diff
@@ -77,7 +80,29 @@ describe("Attestation", () => {
...
       );
-      expect(view.getByRole("textbox")).toHaveValue("123");
+      const input = view.getByLabelText(/Operator Badge Number/i, {
+        selector: "input",
+      });
+      expect(input).toHaveValue("123");
+    });
+  });
+
+  describe("radio text box", () => {
```
skyqrose, line 90: Is there any behavior for the radio box that needs to be tested besides the happy path? Maybe that you can't submit the attestation if it's blank?

skyqrose (approved):
A few small comments, none of them are required and none of them require a re-review unless you want one. 👍

js/components/operatorSignIn/operatorSignInModal.tsx
mathcolo: yeah, I hear you. try again could probably be changed to happen at the network level, but that has other downsides and is a little tricky.

js/test/components/operatorSignIn/attestation.test.tsx
mathcolo: oop, thanks! fixed

PR #166, opened 2024-10-24
fix(operatorSelection): fix Enter/Go key with <form>

mathcolo:
Asana Task: None

In #164 I fixed the Enter/Go key on the Attestation by switching to an HTML `<form>`; doing it here for OperatorSelection too.

Checklist

- Tests:
  - `( )` Has tests
  - `(x)` Doesn't need tests
- Product/Design sign off:
  - `(x)` No review needed

### DIFF
--- a/js/components/operatorSignIn/operatorSelection.tsx
+++ b/js/components/operatorSignIn/operatorSelection.tsx
@@ -42,7 +42,12 @@ export const OperatorSelection = ({
   const buttonEnabled = badgeEntry !== null;
 
   return (
-    <div className="flex flex-col content-stretch">
+    <form
+      className="flex flex-col content-stretch"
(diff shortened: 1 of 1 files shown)


skyqrose (approved):
If this fixes it, 👍 

Another thing that may or may not make a difference:
You could make the button `type=submit` with no `onClick` handler, and then add the JS `onOk()` call to the form's `onSubmit` handler.

I know that's a thing that you can do, but I don't know enough about HTML and the fix you're trying to do to know whether that would actually behave any better/differently.

mathcolo:
Ah, yes, I tried that over in `attestation` in the other PR. Ran into issues because of the React-side button validation. It works here, though, and it's even more semantically correct. Will update and merge 👍

PR #181, opened 2024-11-22
feat: Install Appcues

mathcolo:
Asana Task: [Install Appcues](https://app.asana.com/0/1200273269966439/1207481990454358/f)

Checklist

- Tests:
  - `(x)` Has tests
  - `( )` Doesn't need tests
- Product/Design sign off:
  - `( )` No review needed

### DIFF
--- a/config/dev.secret.example.exs
+++ b/config/dev.secret.example.exs
@@ -4,14 +4,14 @@
 import Config
 
 # Appcues
-config :orbit,
-  appcues_enabled?: true,
-  appcues_id: "SOME_ID"
+# config :orbit,
+#   appcues_enabled?: true,
+#   appcues_id: "SOME_ID"
 
 # FullStory
 # Only turn this on if you're testing sending data to FullStory, not for regular use.
-config :orbit,
-  full_story_org_id: ""
+# config :orbit,
+#   full_story_org_id: ""
 
 # Sentry
 # Only turn this on if you're testing sending events to Sentry, not for regular use.
--- a/config/dev.secret.example.exs
+++ b/config/dev.secret.example.exs
@@ -3,6 +3,11 @@
 # Copy the template to dev.secret.exs, then uncomment and fill in any sections you need.
 import Config
 
+# Appcues
+config :orbit,
+  appcues_enabled?: true,
+  appcues_id: "SOME_ID"
+
 # FullStory
 # Only turn this on if you're testing sending data to FullStory, not for regular use.
 config :orbit,
--- a/config/test.exs
+++ b/config/test.exs
@@ -1,6 +1,8 @@
 import Config
 
 config :orbit,
+  appcues_enabled?: true,
+  appcues_id: "APPKEWES",
   allow_test_data?: true,
   environment: "test",
   release: "test",
--- a/config/test.exs
+++ b/config/test.exs
@@ -1,12 +1,9 @@
 import Config
 
 config :orbit,
-  appcues_enabled?: true,
-  appcues_id: "APPKEWES",
   allow_test_data?: true,
   environment: "test",
   release: "test",
-  full_story_org_id: "RAWR",
   force_https?: false
 
 # Database config
--- a/js/components/appcues.tsx
+++ b/js/components/appcues.tsx
@@ -1,0 +1,45 @@
+import { getMetaContent } from "../util/metadata";
+import { useEffect } from "react";
+import { useLocation } from "react-router-dom";
+
+declare global {
+  const Appcues: {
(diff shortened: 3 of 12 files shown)


config/dev.secret.example.exs
```diff
@@ -3,6 +3,11 @@
 # Copy the template to dev.secret.exs, then uncomment and fill in any sections you need.
 import Config
 
+# Appcues
+config :orbit,
+  appcues_enabled?: true,
+  appcues_id: "SOME_ID"
```
skyqrose, line 9: This should be commented out by default so that copying the template doesn't start sending data to appcues unless you specifically turn it on. (I see that fullstory below is also uncommented but should be commented.)

An alternative would be to leave the `appcues_id` as `""`, and control enabling appcues by whether the id is non-nil and non-empty, instead of having a separate `_enabled?` and `_id` fields. Then it can be safetly left uncommented.

config/test.exs
```diff
@@ -1,6 +1,8 @@
 import Config
 
 config :orbit,
+  appcues_enabled?: true,
+  appcues_id: "APPKEWES",
```
skyqrose, line 5: Will this send data to appcues when running end to end tests with a headless browser?

mathcolo: Ooh, possibly! Changed to use `reassign_env` as we already did for sentry. 06f21df

config/dev.secret.example.exs
mathcolo: 06f21df

js/components/appcues.tsx
```diff
@@ -0,0 +1,45 @@
...
+      properties?: Record<string, string | number | boolean>,
+    ): void;
+    page(): void;
+  };
+}
+
+// React component, put inside react router
+export const AppcuesTrackPage = (): null => {
+  const email = getMetaContent("userEmail") ?? "";
+  const appcuesUserId = getMetaContent("appcuesUserId");
```
skyqrose, line 18: Doesn't `getMetaContent` have an allowlist of known keys? I'm surprised that this PR doesn't have an edit to `js/util/metadata.ts`. Why doesn't that cause typechecking to fail?

lib/orbit_web/controllers/react_app_controller.ex
```diff
@@ -6,9 +6,21 @@ defmodule OrbitWeb.ReactAppController do
...
+      full_story_org_id: Application.get_env(:orbit, :full_story_org_id),
+      sentry_dsn: Application.get_env(:sentry, :dsn)
     )
   end
+
+  # Use half a sha256 of the email address as user id.
+  # This is not secure! But we're eliminating randomness
+  #  while not being trivially overridable.
+  defp appcues_uid(conn) do
+    :crypto.hash(:sha3_256, conn.assigns[:email])
```
skyqrose, line 22: email might be missing (if you haven't logged in I guess?) it's optional in `root.html.heex`. But in that case, we still generate a `uid`. Does that incorrectly imply to fullstory that all logged-out users are the same? Should we avoid generating a uid if there's no email?

js/components/appcues.tsx
skyqrose: Ah, I missed the commit that added that by like a minute.

js/components/appcues.tsx
```diff
@@ -0,0 +1,45 @@
...
+        });
+      } catch (error) {
+        console.warn(error);
+      }
+    }
+  }, [appcuesInitialized, email, appcuesUserId]);
+  useEffect(() => {
+    if (appcuesInitialized) {
+      try {
+        Appcues.page();
```
skyqrose, line 38: Is it important that this happens after `Appcues.identify`? Probably not, but if so, I worry about race conditions and triggering these effects in the wrong order.

skyqrose (approved):

lib/orbit_web/controllers/react_app_controller.ex
mathcolo: Orbit redirects logged-out users to SSO before they can get to `react_app_controller`, so the optional in `root.html.heex` is purely caution (unnecessary?). Both appcues and fullstory will only ever be seeing logged-ins.

js/components/appcues.tsx
mathcolo: I'm not sure :/ this was in response so it being called too often in glides. it's probably fine? https://github.com/mbta/glides/pull/503

mathcolo:
Waiting for https://github.com/mbta/terraform_modules/pull/49 merge before proceeding here

PR #203, opened 2024-12-16
feat: import certifications into database

mathcolo:
Asana Task: [📐 Load Orbit Cert expiries into DB](https://app.asana.com/0/1206105669438487/1208740380342701/f)

Do what Glides does to import certifications, except also include a `rail_line` field based on the certification title.

Checklist

- Tests:
  - `(x)` Has tests
  - `( )` Doesn't need tests
- Product/Design sign off:
  - `( )` No review needed


lib/orbit/import/import_certifications.ex
```diff
@@ -0,0 +1,54 @@
...
+    |> Enum.map(fn certification ->
+      title = certification["Certifications - Certification Title"]
+      expires_string = certification["Certifications - Certification Period Expiration Date"]
+
+      expires =
+        expires_string |> Timex.parse!("{M}/{D}/{YYYY} {h12}:{m} {AM}") |> NaiveDateTime.to_date()
+
+      rail_line =
+        cond do
+          String.ends_with?(title, "BL") ->
```
mathcolo, line 21: @skyqrose heads up, this is a code smell, but was actually mentioned by LMS.

lib/orbit/certification.ex
```diff
@@ -0,0 +1,48 @@
...
+  def changeset(struct, attrs \\ %{}) do
+    struct
+    |> cast(attrs, [
+      :badge,
+      :type,
+      :rail_line,
+      :expires
+    ])
+    |> unique_constraint([:badge, :type])
+    |> validate_required([:expires])
```
skyqrose, line 36: why aren't they all required?

Does `unique_constraint` imply that the fields exist?

lib/orbit/certification.ex
```diff
@@ -0,0 +1,48 @@
...
+
+  def changeset(struct, attrs \\ %{}) do
+    struct
+    |> cast(attrs, [
+      :badge,
+      :type,
+      :rail_line,
+      :expires
+    ])
+    |> unique_constraint([:badge, :type])
```
skyqrose, line 35: Could an employee be certified on multiple `rail_line`s?

priv/repo/migrations/20241211135758_certifications.exs
```diff
@@ -0,0 +1,18 @@
...
+      # Enum
+      add(:type, :integer)
+      # Enum
+      add(:rail_line, :rail_line)
+      add(:expires, :date)
+      timestamps()
+    end
+
+    create unique_index(:certifications, [:badge, :type])
+    create index(:certifications, [:badge])
```
skyqrose, line 16: I think that Postgres can use the prefix of an index. So if there's an index on `[:a, :b, :c]`, then it can also use it as an index on `[:a, :b]`. So this index is redundant with the `unique_index` above.

priv/repo/migrations/20241211135758_certifications.exs
```diff
@@ -0,0 +1,18 @@
...
+      add(:badge, :string)
+      # Enum
+      add(:type, :integer)
+      # Enum
+      add(:rail_line, :rail_line)
+      add(:expires, :date)
+      timestamps()
+    end
+
+    create unique_index(:certifications, [:badge, :type])
```
skyqrose, line 15: Related to the comment in `certification.ex`, should this also have `rail_line` so employees can be certified on more than one line?

lib/orbit/certification.ex
```diff
@@ -0,0 +1,48 @@
...
+    |> validate_required([:expires])
+  end
+
+  @spec get_certifications([String.t()]) :: [%__MODULE__{}]
+  def get_certifications(badges) do
```
skyqrose, line 40: This is unused. `certifications_controller` does its own query directly. I think it should be removed, and the controller can keep its custom query.

lib/orbit/import/import_certifications.ex
```diff
@@ -0,0 +1,54 @@
...
+      }
+    end)
+    |> Enum.group_by(fn certification -> certification.badge end)
+    |> Enum.map(fn {_badge, certifications} ->
+      Enum.max_by(certifications, fn certification -> certification.expires end, Date)
+    end)
+    |> Enum.each(
+      &Repo.insert(&1,
+        on_conflict: :replace_all,
+        conflict_target: [:badge, :type]
```
skyqrose, line 48: (Another place where the unique fields are listed, which would need `rail_line` if we decide to do that.)

lib/orbit/import/import_row_certifications.ex
```diff
@@ -0,0 +1,8 @@
+defmodule Orbit.Import.ImportRoWCertifications do
+  @behaviour Orbit.Import.Importer
+
+  @impl true
+  def import!(file_lines) do
+    Orbit.Import.ImportCertifications.import!(file_lines, :right_of_way)
```
skyqrose, line 6: This needs a whole extra file and layer of indirection in order to pass the `:right_of_way` argument into `ImportCertifications`. Is there a way we could remove that, maybe by having the `import!()` callback take an optional argument (specified in the crontab arguments) so that `ImportCertifications` can be the module that implements the behaviour?

lib/orbit/import/import_certifications.ex
```diff
@@ -0,0 +1,54 @@
...
+  alias Orbit.Certification
+  alias Orbit.Import.ImportHelpers
+  alias Orbit.Repo
+
+  @spec import!(Enumerable.t(), Certification.certification_type()) :: :ok
+  def import!(file_lines, certification_type) do
+    file_lines
+    |> Enum.drop_while(fn line -> String.trim(line) != "" end)
+    |> Enum.drop_while(fn line -> String.trim(line) == "" end)
+    |> ImportHelpers.parse_csv(true)
```
skyqrose, line 11: Can `ImportHelpers.parse_csv` handle these empty line filters? I think we'd want them on all our imports. They might be arguments to the csv library, and if not, then they could be done in our `parse_csv`.

lib/orbit/import/import_certifications.ex
```diff
@@ -0,0 +1,54 @@
...
+            :blue
+
+          String.ends_with?(title, "OL") ->
+            :orange
+
+          String.ends_with?(title, "RL") ->
+            :red
+
+          true ->
+            nil
```
skyqrose, line 31: How do we want to handle lines we can't parse?
- Drop the row
- Crash 
- Save with a missing nil

Currently, this will save the nil, but certification.ex `@type t` doesn't say the field is nillable.

We probably also want to log a warning.

priv/repo/migrations/20241211135758_certifications.exs
```diff
@@ -0,0 +1,18 @@
...
+  use Ecto.Migration
+
+  def change do
+    create table(:certifications) do
+      add(:badge, :string)
+      # Enum
+      add(:type, :integer)
+      # Enum
+      add(:rail_line, :rail_line)
+      add(:expires, :date)
```
skyqrose, line 11: All or most of these fields should probably get `null: false`.

lib/orbit/import/import_certifications.ex
```diff
@@ -0,0 +1,54 @@
...
+      %Certification{
+        badge: String.trim_leading(certification["User - User ID"], "0"),
+        type: certification_type,
+        rail_line: rail_line,
+        expires: expires
+      }
+    end)
+    |> Enum.group_by(fn certification -> certification.badge end)
+    |> Enum.map(fn {_badge, certifications} ->
+      Enum.max_by(certifications, fn certification -> certification.expires end, Date)
```
skyqrose, line 43: (Another more subtle place where this code assumes each employee only has one `rail_line`)

lib/orbit/import/import_helpers.ex
```diff
@@ -0,0 +1,63 @@
+defmodule Orbit.Import.ImportHelpers do
+  @moduledoc """
+  Helpers for importing CSVs from HASTUS
```
skyqrose, line 3: Copy paste oversight: These CSVs aren't from HASTUS

lib/orbit/import/import_helpers.ex
```diff
@@ -0,0 +1,63 @@
...
+        "#{line}\n"
+      end
+    end)
+    |> CSV.decode!(
+      options ++ [headers: headers, field_transform: &String.trim/1, validate_row_length: true]
+    )
+  end
+
+  @spec to_csv(Enumerable.t(), [atom()]) :: [String.t()]
+  def to_csv(data, headers) do
```
skyqrose, line 29: YAGNI, and also this wouldn't be used for importing. Should probably be removed.

lib/orbit/import/import_helpers.ex
```diff
@@ -0,0 +1,63 @@
+defmodule Orbit.Import.ImportHelpers do
```
skyqrose, line 1: This file should probably have tests. It'd mostly be exercised by the tests of the individual importers, but those cases would be complex, so it'd be nice to have some simple tests for these functions, too.

lib/orbit/import/import_certifications.ex
```diff
@@ -0,0 +1,54 @@
+defmodule Orbit.Import.ImportCertifications do
```
skyqrose, line 1: This file is missing tests.

skyqrose (commented):
Wow that was a lot of comments, sorry

lib/orbit/import/async_import.ex
```diff
@@ -0,0 +1,37 @@
+defmodule Orbit.Import.AsyncImport do
+  use Oban.Worker, queue: :cert_import, max_attempts: 1
+
+  alias Orbit.Import.ImportHelpers
+
+  @doc """
+    importer: string name refering to a module in Orbit.Import
+    s3_ref: atom
+    s3_path: string
+    delete: boolean, whether the file should be deleted from s3 after completion
```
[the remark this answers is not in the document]
mathcolo: removed 👍

lib/orbit/certification.ex
mathcolo: Oh, I thought it did—but I looked, and no. fixed! https://github.com/mbta/orbit/pull/203/files#diff-10334beccc32492f2c1511429be01af770e8fd19351dffbde7df02f4c39b9a55R32

lib/orbit/certification.ex
mathcolo: Yep—fixed (same link as above)

priv/repo/migrations/20241211135758_certifications.exs
mathcolo: Fixed https://github.com/mbta/orbit/pull/203/files#diff-248c4b179027894763ea37f20c43c9e2ac96ea5b22619c4e898777d809a5c8b5R15

lib/orbit/import/import_certifications.ex
mathcolo: https://github.com/mbta/orbit/pull/203/files#diff-5a8c606e5cebbccd139908748cdfb09e3ad72c2ad41757acb8373c22c4c2a9beR64

lib/orbit/import/import_certifications.ex
mathcolo: These are actually explicitly for the LMS csvs that contain custom headers at the top—it's dropping those and the empty lines that precede the actual csv data. They aren't csv specific and should remain outside of `ImportHelpers`.

priv/repo/migrations/20241211135758_certifications.exs
mathcolo: Agreed, fixed https://github.com/mbta/orbit/pull/203/files#diff-248c4b179027894763ea37f20c43c9e2ac96ea5b22619c4e898777d809a5c8b5R6

lib/orbit/import/import_certifications.ex
mathcolo: 🤦 thanks; fixed https://github.com/mbta/orbit/pull/203/files#diff-5a8c606e5cebbccd139908748cdfb09e3ad72c2ad41757acb8373c22c4c2a9beR57

lib/orbit/import/import_certifications.ex
mathcolo: 🤦 https://github.com/mbta/orbit/pull/203/files#diff-9d41583cd5e179dab2b68a3c23d935102e1b9c49e3a66831fb668f314c0c7ecdR1

lib/orbit/import/scheduled_row_certification_import.ex
```diff
@@ -0,0 +1,19 @@
...
+  alias Orbit.Import.AsyncImport
+
+  @impl Oban.Worker
+  def perform(_) do
+    %{
+      importer: "right_of_way_certifications",
+      s3_ref: :glides_global,
+      s3_path: "cornerstone/right_of_way_certification.csv",
+      delete: false
+    }
```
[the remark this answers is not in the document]
mathcolo: Refactored to use Oban `args` and remove the behaviour, but I avoided touching Rfid and Personnel for now https://github.com/mbta/orbit/pull/203/files#diff-8e72c709985cb6a43cc0ab32cb487b8c17d2b2d003338ea42a4e2ad2829979dcR27

lib/orbit/import/import_helpers.ex
mathcolo: yeah, this was a copypasta lesson. it's never as simple as it looks to do that!

lib/orbit/import/import_helpers.ex
mathcolo: yeah, you're right. here are a couple for `parse_csv`—the other two functions are really just wrappers? https://github.com/mbta/orbit/pull/203/files#diff-b5202e24be21249b1155432d249572dd4a4b0a3b844f445fe4051215d31579deR1

priv/repo/migrations/20241211135758_certifications.exs
mathcolo: yes 👍 good callout.... https://github.com/mbta/orbit/pull/203/files#diff-248c4b179027894763ea37f20c43c9e2ac96ea5b22619c4e898777d809a5c8b5R15

lib/orbit/import/import_certifications.ex
mathcolo: Changed to be nillable; uncertain about the warning. We aren't generally logging these sorts of things, so they're not useful if we aren't looking for them. Are you thinking we would throw a splunk alert on it?

mathcolo:
@skyqrose Let me know what you think; I threw everything (except for the configuration) into `Orbit.Import.ImportCertifications`, which is clean and a lot easier to follow now. Also fixed the remaining items (🤦); thanks for looking!

lib/orbit/certification.ex
```diff
@@ -0,0 +1,35 @@
...
+
+  @type t :: %__MODULE__{
+          badge: String.t(),
+          type: certification_type(),
+          rail_line: RailLine.t() | nil,
+          expires: Date.t()
+        }
+  schema "certifications" do
+    field(:badge, :string)
+    field(:type, Ecto.Enum, values: [right_of_way: 0, rail: 1])
```
skyqrose, line 18: Most places we use `Ecto.Enum`, we save the values as atoms (which get converted to strings in the database), though I see the Glides code this is copied from uses ints (it's the only place that uses ints).

Have you thought about whether you'd rather use ints or atoms/strings here?

lib/orbit/import/import_certifications.ex
skyqrose: Definitely not worth a splunk alert. I was thinking it might occasionally be useful if you see it locally or if doing a manual search for it while debugging an issue.

But I guess if we're manually debugging something, it's just as easy to grab the CSV from S3 and see the missing title there.

If we expect this to show up rarely or never, then might as well log it, that's what Log.warn is for, but it's definitely not necessary.

lib/orbit/import/helpers.ex
```diff
@@ -0,0 +1,43 @@
...
+  Read the csv from the path
+  """
+  @spec import_from_file!(String.t()) :: File.Stream.t()
+  def import_from_file!(file_path) do
+    file_path
+    |> File.stream!()
+  end
+
+  @doc """
+  Read the csv from the `:glides` S3 ref in the given path, and pass it to the provided importer
```
skyqrose, line 34: nit: What does `` `:glides` `` mean? I think it's a copy-paste from Glides, and might be a reference to how the refs are specified in config, but I don't think it's relevant here.

```suggestion
  Read the csv from the S3 ref at the given path, and pass it to the provided importer
```

skyqrose (approved):
Wow this cleaned up. I think it was 17 files before, and now it's 13, including the new tests.

Looks good.

lib/orbit/certification.ex
mathcolo: I haven't, but I am now... 

Although I admit integers don't make a ton of sense for values that don't have an ordering, I suppose I'm not sure using a physically wide space for an enum makes more sense either :( Lose/lose, going to leave as in glides.

PR #205, opened 2024-12-20
feat: frontend expiring certification warnings and sign-in prevention

mathcolo:
Asana Task: [📐 Implement Warnings for Expiring Certifications](https://app.asana.com/0/1206105669438487/1208729753338210/f)

image
image
image
image

Checklist

- https://github.com/mbta/terraform_modules/pull/58
- Tests:
  - `(x)` Has tests
  - `( )` Doesn't need tests
- Product/Design sign off:
  - `( )` No review needed


js/test/components/operatorSignIn/attestation.test.tsx
```diff
@@ -216,4 +227,42 @@ describe("Attestation", () => {
...
+        />,
+      );
+      await userEvent.click(
+        view.getByRole("button", { name: "Continue to Fit for Duty Check →" }),
+      );
+      expect(
+        view.getByRole("button", { name: "Complete Fit for Duty Check" }),
+      ).toBeInTheDocument();
+    });
+  });
```
skyqrose, line 267: these tests are redundant. You don't need to assert the existence of the button and then assert that you can click it, just one test that asserts clicking the button would work.

I suggest deleting the first test, and copying the one unique assertion from that test (`getByText("ExpiredCard")`) into the second.
```suggestion
    test("can bypass expired mode", async () => {
      const view = render(
        <Attestation
          badge="123"
          prefill={false}
          onComplete={jest.fn()}
          loading={false}
          employees={EMPLOYEES}
          certifications={CERTIFICATIONS_ONE_EXPIRED}
        />,
      );
      expect(view.getByText("Expired card")).toBeInTheDocument();
      await userEvent.click(
        view.getByRole("button", { name: "Continue to Fit for Duty Check →" }),
      );
      expect(
        view.getByRole("button", { name: "Complete Fit for Duty Check" }),
      ).toBeInTheDocument();
    });
  });
```

js/test/components/operatorSignIn/attestation.test.tsx
```diff
@@ -216,4 +227,42 @@ describe("Attestation", () => {
     expect(onComplete).not.toHaveBeenCalled();
     jest.useRealTimers();
   });
+
+  describe("expiry", () => {
```
skyqrose, line 231: `attestation.test.tsx` now has a test for the expired case. But does it need a test that it shows the about to expire warning? I see that the details of that situation are asserted in `expiry.test.tsx`, but I'm thinking about a test that proves that `CertificationBoxes` is called.

js/components/operatorSignIn/expiry.tsx
```diff
@@ -0,0 +1,163 @@
...
+            now,
+            c.expires,
+          )} on ${c.expires.toLocaleString(DateTime.DATE_SHORT)}`,
+      )
+      .join(" and ") +
+    ".";
+  return (
+    <WarningParagraph
+      className={className([
+        "border-0 light:bg-opacity-40 dark:bg-opacity-30 dark:text-white",
```
skyqrose, line 58: refactoring suggestion: This is the only use of `WarningParagraph`, and these styles don't need to be controlled dynamically, so should they be part of `WarningParagraph` instead of passed in as an argument?

And should the background color be controlled by passing in `"warning"` or `"error"` instead of by giving an extraClassName?

Or, alternatively, WarningParagraph is just 2 divs and some styles, and it's not used anywhere else, so maybe it should be inlined here instead of abstracted?

js/components/operatorSignIn/expiry.tsx
```diff
@@ -0,0 +1,163 @@
...
+          }
+          operatorName={displayName}
+          certifications={expiresSoon}
+        />
+      )}
+      {expired.length > 0 && !ignoreExpired && (
+        <CertificateBox
+          now={now}
+          mode="error"
+          title={expired.length === 2 ? "Expired cards" : "Expired card"}
```
skyqrose, line 105: nitpick: This is fragile if anybody ever has more than 2 certifications (could happen if we track more than two types, or if someone has certs on more than one rail line). It should be `expired.length === 1 ? singular : plural` or `expired.length > 1 ? plural : singular`.

Also applies to the previous call 10 lines above this.

js/components/operatorSignIn/expiry.tsx
```diff
@@ -0,0 +1,163 @@
...
+      <div className="m-2 text-sm text-gray-400">
+        <p>Is this warning incorrect?</p>
+        <span>
+          If {displayName} has a valid{" "}
+          {expireds.map((c) => humanReadableType(c.type)).join(" and ")}:
+        </span>
+        <ol className="mb-4 ml-10 mr-0 mt-2 list-decimal">
+          <li className="m-0">
+            Take a picture of{" "}
+            {expireds.length === 2 ? "both cards" : "the card"}.
```
skyqrose, line 152: What if length > 2?

js/components/operatorSignIn/expiry.tsx
```diff
@@ -0,0 +1,163 @@
...
+  certifications,
+  now,
+  onContinue,
+}: {
+  displayName: string;
+  certifications: Certification[];
+  now: DateTime;
+  onContinue: () => void;
+}): ReactElement => {
+  const expireds = certifications.filter((c) => isExpired(c, now));
```
skyqrose, line 138: I don't think this would actually cause problems, but it seems like we shouldn't have to recalculate which cards are expired at this point. The calculations are all in sync now but what if something changes? Perhaps the caller should calculate `expireds` once, and use the same value both for deciding whether to show `Bypass` and for passing in as an argument:

```tsx
expireds = certifications.filter((c) => isExpired(c, now));
...
(expireds.length > 0) ? <Bypass expireds={expireds} displayName={...} onContinue={...} />
```

js/test/components/operatorSignIn/operatorSignInModal.test.tsx
```diff
@@ -161,7 +176,6 @@ describe("OperatorSignInModal", () => {
     await userEvent.type(view.getByRole("textbox"), "123");
     await userEvent.click(view.getByRole("button", { name: "OK" }));
 
-    expect(view.getByText("Step 2 of 2")).toBeInTheDocument();
```
skyqrose, line 164: Why are these assertions removed?

skyqrose (commented):
There was some talk about whether we should warn on someone not having any certifications in the database. I don't remember the resolution of that discussion. Should anything like that be incorporated into this feature?

js/hooks/useCertifications.ts
```diff
@@ -0,0 +1,19 @@
...
+  certificationFromData,
+} from "../models/certification";
+
+const CERTIFICATIONS_API_PATH = "/api/certifications";
+
+const parse = (list: CertificationDataList) => {
+  return list.map(certificationFromData);
+};
+
+export const useCertifications = (badge: string | null) => {
```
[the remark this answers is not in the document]
mathcolo: yep! [fixed](https://github.com/mbta/orbit/pull/205/files#diff-17217e575910889ec3ff8935b992dfffae1185bd069d13b4511217795e8e977dR16)

js/models/certification.ts
```diff
@@ -0,0 +1,55 @@
+import { DateTime } from "luxon";
+import { z } from "zod";
+
+export type Certification = {
+  type: string;
+  rail_line: string;
```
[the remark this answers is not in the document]
mathcolo: it should be; thanks

js/test/components/operatorSignIn/attestation.test.tsx
mathcolo: fixed 👍

js/components/operatorSignIn/expiry.tsx
mathcolo: Inlined 👍 yeah, I honestly don't know why I made this separate in glides (which I copypastaed over here!)

js/test/components/operatorSignIn/operatorSignInModal.test.tsx
mathcolo: Per [the task AC](https://app.asana.com/0/1206105669438487/1208729753338210/f) we're removing the "Step 2 of 2" language

js/components/operatorSignIn/expiry.tsx
mathcolo: fixed 👍 thanks

js/components/operatorSignIn/expiry.tsx
mathcolo: fixed 👍 thanks

js/components/operatorSignIn/attestation.tsx
```diff
@@ -1,90 +1,100 @@
...
+                onComplete(enteredRadio);
+              }}
+              disabled={!valid}
+            >
+              Complete Fit for Duty Check
+            </button>
+          </form>
+        </>
+      : <>
+          <Instructions displayName={name} />
```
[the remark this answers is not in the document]
mathcolo: inlined 👍

js/components/operatorSignIn/expiry.tsx
```diff
@@ -0,0 +1,147 @@
...
+          )} on ${c.expires.toLocaleString(DateTime.DATE_SHORT)}`,
+      )
+      .join(" and ") +
+    ".";
+  return (
+    <div
+      className={className([
+        "mb-4 mt-2 flex flex-row rounded border-0 px-3 py-2",
+        mode === "warning" && "bg-[#FFDE9E]",
+        mode === "error" && "bg-[#FF919A]",
```
skyqrose, line 59: A previous version (when there was the WarningParagraph component) had some extra classes. `light:bg-opacity-40 dark:bg-opacity-30 dark:text-white`. Did you mean to remove those in f54a72135dc4800702ef9b13322783a12595e4e1 ? (Also, I don't think `border-0` is needed anymore, since no border should be the default.)

lib/orbit_web/controllers/certifications_controller.ex
```diff
@@ -20,5 +20,19 @@ defmodule OrbitWeb.CertificationsController do
...
+    #       rail_line: "blue",
+    #       expires: "2024-12-24"
+    #     },
+    #     %{
+    #       type: "right_of_way",
+    #       rail_line: "blue",
+    #       expires: "2025-01-24"
+    #     }
+    #   ]
+    # })
```
skyqrose, line 36: Left some dead code / test data in here.

skyqrose (approved):
A couple comments, but no re-review needed after.

js/components/operatorSignIn/expiry.tsx
mathcolo: Indeed, all intentional 👍 we don't have light/dark mode and the colors are exact. thanks for catching border-0!

PR #218, opened 2025-01-07
fix csp warning, typo and allow react developer tools

skyqrose:
Asana Task: None, noticed while working on [Orbit auto-refresh](https://app.asana.com/0/1200689710863995/1208671637680878/f)

**Before**:
Screenshot 2025-01-07 at 16 18 22
and the React Developer Tools extension didn't work.

**After**:
The cookie warning is still there, but the others are gone, and react developer tools works.

It is relaxing security a bit to allow `script-src-elem 'unsafe-inline'` but unfortunately I couldn't find a more specific way to allow the extension to work. And this extension is super useful to me, so I think it's worth the security-convenience tradeoff.

React Developer Tools still gave another warning about not being able to find its source maps. I couldn't find how to allow that in the CSP, and seeing the extension's source maps isn't important, so if you use this extension then I recommend turning off "show inline warnings" in its settings:

Screenshot 2025-01-07 at 16 20 55

- Tests:
  - `( )` Has tests
  - `(x)` Doesn't need tests
- Product/Design sign off:
  - `(x)` No review needed

### DIFF
--- a/lib/orbit_web/router.ex
+++ b/lib/orbit_web/router.ex
@@ -15,13 +15,15 @@ defmodule OrbitWeb.Router do
 
     plug :put_secure_browser_headers, %{
       # much of this is necessary for Appcues: https://docs.appcues.com/user-experiences-faq/faq-content-security-policies
+      # script-src-elem 'unsafe-inline' is for react developer tools
       "content-security-policy" => "\
         connect-src 'self' *.sentry.io *.fullstory.com https://*.appcues.com https://*.appcues.net wss://*.appcues.net wss://*.appcues.com;\
         default-src 'self';\
-        font-src self' https://fonts.gstatic.com;\
+        font-src 'self' https://fonts.gstatic.com;\
         frame-src 'self' https://*.appcues.com;\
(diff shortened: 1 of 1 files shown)


lib/orbit_web/router.ex
```diff
@@ -15,13 +15,15 @@ defmodule OrbitWeb.Router do
 
     plug :put_secure_browser_headers, %{
       # much of this is necessary for Appcues: https://docs.appcues.com/user-experiences-faq/faq-content-security-policies
+      # script-src-elem 'unsafe-inline' is for react developer tools
```
mathcolo, line 18: Can this comment go above 26, or does `mix format` dislike that?

mathcolo (approved):

lib/orbit_web/router.ex
skyqrose: content security policy is one long string from line 19 to 28 (note the backslashes), so it's syntactic issue that you can't put a comment in the middle of a string, not a formatting issue.

PR #219, opened 2025-01-07
fix(certifications): filter out certifications that aren't relevant for BL signin

mathcolo:
Asana Task: [📐 Implement Warnings for Expiring Certifications](https://app.asana.com/0/1206105669438487/1208729753338210/f)

Mix of a feature and a fix. Filter out certifications for other lines, so that:
1. (feature) we only evaluate BL operators against BL certifications. will make RL and OL easier.
2. (fix) old rail certifications in the data that are lineless (don't match the format) are filtered out.

Checklist

- Tests:
  - `(x)` Has tests
  - `( )` Doesn't need tests
- Product/Design sign off:
  - `( )` No review needed

### DIFF
--- a/js/models/certification.ts
+++ b/js/models/certification.ts
@@ -1,15 +1,16 @@
+import { HeavyRailLine } from "../types";
 import { DateTime } from "luxon";
 import { z } from "zod";
 
 export type Certification = {
   type: string;
-  railLine: string;
+  railLine: HeavyRailLine;
   expires: DateTime;
 };
 
 export const CertificationData = z.object({
   type: z.string(),
-  rail_line: z.string(),
+  rail_line: HeavyRailLine,
   expires: z.string(),
 });
 export type CertificationData = z.infer<typeof CertificationData>;
@@ -29,7 +30,20 @@ export const isExpired = (c: Certification, now: DateTime) => {
   return now >= c.expires.startOf("day");
 };
 
-export const getExpired = (cs: Certification[] | undefined, now: DateTime) => {
+export const filterRelevantForOperator = (
+  cs: Certification[],
+  line: HeavyRailLine,
+) => {
+  return cs.filter(
+    (c) =>
+      (c.type === "rail" && c.railLine === line) || c.type === "right_of_way",
+  );
+};
+
+export const filterExpired = (
+  cs: Certification[] | undefined,
+  now: DateTime,
+) => {
   if (cs === undefined) {
     return [];
   }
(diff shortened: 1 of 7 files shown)


js/models/certification.ts
```diff
@@ -29,7 +30,20 @@ export const isExpired = (c: Certification, now: DateTime) => {
...
 };
 
-export const getExpired = (cs: Certification[] | undefined, now: DateTime) => {
+export const filterRelevantForOperators = (
+  cs: Certification[],
+  line: HeavyRailLine,
+) => {
+  return cs.filter(
+    (c) =>
+      (c.type === "rail" && c.railLine === line) || c.type === "right_of_way",
```
skyqrose, line 39: Do ROW certs have a line associated with them? Can you use an OL ROW cert on the BL? (I don't think they have separate ROW trainings by line, so I think this is working correctly, but wanna double check.)

Also there's no test for whether an OL ROW cert can apply to BL (all other cases are covered tho, and if you think that test isn't needed that's fine.)

js/test/hooks/useCertifications.test.ts
```diff
@@ -1,16 +1,25 @@
...
 
 jest.mock("../../browser", () => ({
   fetch: jest.fn(),
 }));
 
+// eslint-disable-next-line @typescript-eslint/no-unsafe-return
+jest.mock("../../models/certification", () => ({
+  ...jest.requireActual("../../models/certification"),
+  // eslint-disable-next-line @typescript-eslint/no-unsafe-return
+  filterRelevantForOperators: jest.fn((cs, _line) => cs),
```
skyqrose, line 20: This is a pure function, so you shouldn't need to mock it or assert that it was called.

You could instead use test data that includes a BL rail cert and a OL rail cert, and assert that the hook only returns the BL cert. (This would probably be easier to do in the existing test case rather than writing a second case.)

skyqrose (approved):

js/models/certification.ts
mathcolo: ROW certs are line agnostic, that's correct, so that case is really undefined behavior. _However_, because I tested the hypothetical case whether a blue ROW cert exists, I should probably test the case where an unmatching-line ROW cert exists and is ignored.

js/test/hooks/useCertifications.test.ts
mathcolo: 6137986

I was doing the mock originally to prevent the helper from getting called even though it's pure (so if its behavior changed the test would still pass), but I admit that isn't strictly necessary. Changed!

PR #220, opened 2025-01-08
feat: store FFD override

mathcolo:
Asana Task: [📐 If Orbit FFD expiry is overruled, store that and export it](https://app.asana.com/0/1206105669438487/1208740380342704/f)

- Rename `bypass` key in API request to `override` to match Glides
- Add new `override` column
- Save `override` to database if it passes some rudimentary checks
- Add data to export CSV

Checklist

- Tests:
  - `(x)` Has tests
  - `( )` Doesn't need tests
- Product/Design sign off:
  - `( )` No review needed

### DIFF
--- a/js/models/certification.ts
+++ b/js/models/certification.ts
@@ -1,15 +1,16 @@
+import { HeavyRailLine } from "../types";
 import { DateTime } from "luxon";
 import { z } from "zod";
 
 export type Certification = {
   type: string;
-  railLine: string;
+  railLine: HeavyRailLine;
   expires: DateTime;
 };
 
 export const CertificationData = z.object({
   type: z.string(),
-  rail_line: z.string(),
+  rail_line: HeavyRailLine,
   expires: z.string(),
 });
 export type CertificationData = z.infer<typeof CertificationData>;
@@ -45,7 +46,21 @@ export const isExpired = (c: Certification, now: DateTime) => {
   return now >= c.expires.startOf("day");
 };
 
-export const getExpired = (cs: Certification[] | undefined, now: DateTime) => {
+export const filterRelevantForOperators = (
+  cs: Certification[],
+  line: HeavyRailLine,
+) => {
+  return cs.filter(
+    (c) =>
+      (c.type === "rail" && c.railLine === line) || // Rail cert: line must match
+      (c.type === "right_of_way" && [line, "none"].includes(c.railLine)), // ROW cert: must be 'none' or same line
+  );
+};
+
+export const filterExpired = (
+  cs: Certification[] | undefined,
+  now: DateTime,
+) => {
   if (cs === undefined) {
     return [];
   }
@@ -56,7 +71,7 @@ export const anyOfExpired = (
   cs: Certification[] | undefined,
   now: DateTime,
 ) => {
-  return getExpired(cs, now).length > 0;
+  return filterExpired(cs, now).length > 0;
 };
 
 export const humanReadableType = (type: string): string => {
--- a/js/models/certification.ts
+++ b/js/models/certification.ts
@@ -1,3 +1,4 @@
+import { dateTimeFromISO, dateTimeToISODate } from "../dateTime";
 import { HeavyRailLine } from "../types";
 import { DateTime } from "luxon";
 import { z } from "zod";
@@ -19,22 +20,16 @@ export const CertificationDataList = z.array(CertificationData);
 export type CertificationDataList = z.infer<typeof CertificationDataList>;
 
 export const certificationFromData = (cd: CertificationData): Certification => {
+  const expires = dateTimeFromISO(cd.expires);
   return {
     type: cd.type,
     railLine: cd.rail_line,
-    expires: DateTime.fromISO(cd.expires, { zone: "America/New_York" }),
+    expires,
   };
 };
 
 export const certificationToData = (c: Certification): CertificationData => {
-  const expires = c.expires.toISO();
-  if (expires === null) {
-    // Unclear why luxon's `toISO` could return null, but realistically that
-    //  should never happen
-    throw new Error(
-      "Cannot transform Certification to CertificationData; toISO failed",
-    );
-  }
+  const expires = dateTimeToISODate(c.expires);
   return {
     type: c.type,
     rail_line: c.railLine,
--- a/js/models/certification.ts
+++ b/js/models/certification.ts
@@ -26,10 +26,18 @@ export const certificationFromData = (cd: CertificationData): Certification => {
 };
 
 export const certificationToData = (c: Certification): CertificationData => {
+  const expires = c.expires.toISO();
+  if (expires === null) {
+    // Unclear why luxon's `toISO` could return null, but realistically that
+    //  should never happen
+    throw new Error(
+      "Cannot transform Certification to CertificationData; toISO failed",
+    );
+  }
   return {
     type: c.type,
     rail_line: c.railLine,
-    expires: c.expires.toISO() ?? "TODO why would this be null...",
+    expires,
   };
 };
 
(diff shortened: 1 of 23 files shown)


lib/orbit_web/controllers/sign_in_export_controller.ex
```diff
@@ -31,6 +31,18 @@ defmodule OrbitWeb.SignInExportController do
...
+  @spec override_expiration_date([map()], String.t()) :: String.t()
+  def override_expiration_date(override_list, type) do
+    {:ok, datetime, _} =
+      Enum.find(override_list, fn cert -> cert["type"] == type end)["expires"]
+      |> DateTime.from_iso8601()
+
+    datetime
+    |> DateTime.shift_zone!("America/New_York")
+    |> DateTime.to_date()
+    |> Date.to_iso8601()
```
mathcolo, line 43: Heads up- this is a little messy because the `expires` is stored in the jsonb column as an ISO timestamp, but I want to provide Excel users with an Excel-compatible date in `YYYY-MM-DD`. (Yes, this is lossy)

js/test/components/operatorSignIn/operatorSignInModal.test.tsx
```diff
@@ -122,6 +130,54 @@ describe("OperatorSignInModal", () => {
...
+    } as Response);
+
+    const view = render(
+      <OperatorSignInModal
+        show={true}
+        onComplete={jest.fn()}
+        close={jest.fn()}
+      />,
+    );
+    await userEvent.type(view.getByRole("textbox"), "1234");
```
skyqrose, line 155: naming nitpick:

Reading this test it took me a long time to figure out how it mocks the expired cert. Now I see it's based on the badge number, "123" is certified and "1234" has an expired cert.

I think it'd be easier to follow the test if the operator had a badge string like `"expiredcert"` or something. Or, alternatively, a comment on this line explaining that this badge number is controlling which cert to get, so that anybody editing this test in the future can tell this string is important.

lib/orbit_web/controllers/sign_in_controller.ex
```diff
@@ -61,33 +62,51 @@ defmodule OrbitWeb.SignInController do
     )
   end
 
+  @spec validate_override(map()) :: boolean()
+  defp validate_override(override) do
```
skyqrose, line 66: refactoring suggestion (but probably not): I wish I knew more about ecto changeset validation, because this is the textbook use case for it (taking form input, validating and immediately inserting it to the database, and returning an user-facing error if it's invalid).

If you were motivated to do this in an Ecto-y way, you could try rewriting this to use Ecto validation. But this vanilla elixir validation is perfectly fine and in line with the rest of the app, I've never really understood Ecto's mental model so I can't suggest how to do it, and it's probably not worth the effort to refactor it in an already-complex PR.

test/orbit_web/controllers/sign_in_controller_test.exs
```diff
@@ -200,5 +214,48 @@ defmodule OrbitWeb.SignInControllerTest do
...
+              "type" => "NOT_A_REAL_TYPE"
+            }
+          ]
+        })
+
+      assert "Invalid override\n" = response(conn, 400)
+    end
+
+    @tag :authenticated
+    test "replies 400 when invalid override (extraneous key)", %{conn: conn} do
```
skyqrose, line 240: Why is this important to verify in a test? Do we care if the server ignores vs 400s on an extra key?

lib/orbit_web/controllers/sign_in_export_controller.ex
```diff
@@ -31,6 +31,18 @@ defmodule OrbitWeb.SignInExportController do
     end
   end
 
+  @spec override_expiration_date([map()], String.t()) :: String.t()
+  def override_expiration_date(override_list, type) do
+    {:ok, datetime, _} =
+      Enum.find(override_list, fn cert -> cert["type"] == type end)["expires"]
```
skyqrose, line 37: bug: This will crash if given a type that's not in the list. But below, if there are any overrides, then this is called with both types, even if only one type was overridden.

This should gracefully return `""` or `nil` if the type isn't found.

lib/orbit_web/controllers/sign_in_export_controller.ex
skyqrose: Should we always work with expiration dates as Dates / `YYYY-MM-DD` strings instead of converting them to datetimes then? Or at least save them in the database like that?

I think the date better represents the data. Especially since in datetimes, "Jan 13" is usually represented as `2025-01-13 00:00:00`, but expiration dates usually mean expiring at the end of the day, so at `2025-01-13 23:59:59`, so the datetime isn't even the right datetime.

skyqrose (commented):
Sorry for the delayed review.

lib/util/map.ex
```diff
@@ -0,0 +1,9 @@
+defmodule Util.Map do
+  @spec has_only(map(), MapSet.t()) :: boolean()
+  def has_only(map, keys) do
```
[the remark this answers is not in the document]
mathcolo: agreed 👍 https://github.com/mbta/orbit/pull/220/files#diff-c6b39fbd13b0b85c8a2e45ed1702638d401f59ca6f0a914450256488d87bd0c2R2

js/models/certification.ts
```diff
@@ -26,6 +26,22 @@ export const certificationFromData = (cd: CertificationData): Certification => {
   };
 };
 
+export const certificationToData = (c: Certification): CertificationData => {
+  const expires = c.expires.toISO();
+  if (expires === null) {
+    // Unclear why luxon's `toISO` could return null, but realistically that
```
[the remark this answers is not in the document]
mathcolo: ah, yeah, it looks like it's forcing you to handle at compile time whether it's valid or not...generally. It seems to be happy when the functions are abstracted away to [dateTime.ts](https://github.com/mbta/orbit/pull/220/files#diff-b42da304cf6a0bb10ace57b54df1baf6412aa119e5c8c8a8d84dea7f510708cfR11), and the conversion to string takes in a `<true>`.

js/test/components/operatorSignIn/operatorSignInModal.test.tsx
mathcolo: added a comment above it 👍 https://github.com/mbta/orbit/pull/220/files#diff-cb55066e3d310eb5b472f93dd2e0ff0f71aa20d9b3b346c5a7bb8b40f25b68d5R155

test/orbit_web/controllers/sign_in_controller_test.exs
mathcolo: It seemed like extraneous keys shouldn't be allowed, since that suggests there's something amiss. It's like homebrew reverse zod. Although I admit that might be overly strict and prone to issues. IDK, does that change your leaning?

lib/orbit_web/controllers/sign_in_export_controller.ex
mathcolo: [fixed](https://github.com/mbta/orbit/pull/220/files#diff-10e30af09b064503abd15c28480b0de62aaab8694e92cb3af201ae79a85662d1R36), thanks! using "" as it needs to CSV happy.

lib/orbit_web/controllers/sign_in_controller.ex
mathcolo: Ecto.Changeset never even occurred to me here, but I see it too. I know almost nothing about it. This would be good tech debt for us to be forced to learn broadly; I'll make an inbound task

mathcolo: https://app.asana.com/0/1200882337457260/1209182581824201/f

PR #225, opened 2025-01-13
connect to socket and reload if out of date or unauthenticated

skyqrose:
Asana Task: [📐 Orbit auto-refresh](https://app.asana.com/0/1200689710863995/1208671637680878/f)

Status: I'm not quite done, but I wanted to get what I have pushed, and I'm ready for reviews.

Things I still need to do:
- [x] Frontend tests
- [x] Check that it works on sandbox (label is applied)
- [ ] One more round of manual testing

Makes a new socket. The server tells the client if they're unauthenticated and need to refresh to re-login, or out of date and need to refresh to get a new version.

Unfortunately you can't give an error reason when rejecting a socket connection, so it has to accept the connection and then send the reply. And also unfortunately, you can't reply to a socket connection, so the client has to join a channel and get the reply to that channel join. (I made a phoenix feature request to allow replies so we wouldn't need those workarounds: [Elixir Forum: Feature request: Reply to socket connect](https://elixirforum.com/t/feature-request-reply-to-socket-connect/68694))

- Tests:
  - `( )` Has tests
  - `( )` Doesn't need tests
- Product/Design sign off:
  - `( )` No review needed

### DIFF
--- a/js/app.tsx
+++ b/js/app.tsx
@@ -2,7 +2,10 @@
 import "./telemetry";
 import "./phoenix.js";
 import { App } from "./components/app.js";
+import { initSocket } from "./socket";
 import { createRoot } from "react-dom/client";
+
+initSocket();
 
 // eslint-disable-next-line @typescript-eslint/no-non-null-assertion
 const root = createRoot(document.getElementById("app")!);
--- a/js/socket.ts
+++ b/js/socket.ts
@@ -40,8 +40,8 @@ export const initSocket = () => {
         console.warn(response);
       })
       .receive("error", (response: unknown) => {
-        console.warn("channel error");
-        console.warn(response);
+        console.error("channel error");
+        console.error(response);
       });
   });
 
--- a/js/socket.ts
+++ b/js/socket.ts
@@ -1,0 +1,54 @@
+import { reload } from "./browser";
+import { getMetaContent } from "./util/metadata";
+import { Socket } from "phoenix";
+import { z } from "zod";
+
+export const initSocket = () => {
+  const clientRelease = getMetaContent("release");
+  const socket = new Socket("/socket", {
+    params: {
+      release: clientRelease,
+      token: getMetaContent("guardianToken"),
+    },
+  });
+
+  socket.onError((_error, _transport, _establishedConnections) => {
+    console.warn("socket error");
+  });
+
+  socket.onOpen(() => {
+    socket
+      .channel("metadata")
+      .join()
+      .receive("ok", (rawResponse: unknown) => {
+        const { authenticated, server_release: serverRelease } =
+          MetadataJoinResponse.parse(rawResponse);
+        if (serverRelease !== clientRelease) {
+          console.warn(
+            `Server has version ${serverRelease} but we have version ${clientRelease}. Reloading.`,
+          );
+          reload();
+        }
+        if (!authenticated) {
+          console.warn("Authentication failed. Reloading.");
+          reload();
+        }
+        // channel successfully joined
+      })
+      .receive("timeout", (response: unknown) => {
+        console.warn("channel timeout");
+        console.warn(response);
+      })
+      .receive("error", (response: unknown) => {
+        console.warn("channel error");
+        console.warn(response);
+      });
+  });
+
+  socket.connect();
+};
+
+const MetadataJoinResponse = z.object({
+  authenticated: z.boolean(),
+  server_release: z.string(),
+});
--- a/lib/orbit_web/channels/user_socket.ex
+++ b/lib/orbit_web/channels/user_socket.ex
@@ -1,0 +1,94 @@
+defmodule OrbitWeb.UserSocket do
+  @moduledoc """
+  Will always allow connection, even if user is unauthenticated or out of date,
+  So that we can tell the client what the issue is and they can handle it (by refreshing).
+
+  If we refuse unauthenticated connections, then the client can't tell the difference between
+  an auth problem where they should log in again, and a network problem where they should reconnect.
+  Phoenix has a way to reply with a connection error, and it reaches the browser, but not in a
+  way that's visible to the frontend application code.
+
+  After connection, clients should connect to the `metadata` channel to learn whether they were
+  authenticated or not.
+
+  All other channels should remember to check the release version and
+  authentication/authorization before allowing a client to join the channel.
+  """
+
+  use Phoenix.Socket
+  require Logger
+
+  channel "metadata", OrbitWeb.MetadataChannel
+
+  @doc """
+  fields that are added to socket.assigns
+  """
+  @type metadata :: %{
+          server_release: String.t(),
+          client_release: String.t() | nil,
+          release_matches: boolean(),
+          authenticated: boolean(),
+          user_id: integer() | nil,
+          email: String.t() | nil,
+          authentication_error: :no_token | :invalid_token | nil
+        }
+
+  @impl true
+  def connect(params, socket, _connect_info) do
+    server_release = Application.fetch_env!(:orbit, :release)
+    client_release = params["release"]
+
+    {socket, user, authentication_error} =
+      case Guardian.Phoenix.Socket.authenticate(socket, OrbitWeb.Auth.Guardian, params["token"]) do
+        {:error, :no_token} ->
+          {socket, nil, :no_token}
+
+        {:error, _e} ->
+          {socket, nil, :invalid_token}
+
+        {:ok, authed_socket} ->
+          user = Guardian.Phoenix.Socket.current_resource(authed_socket)
+          {authed_socket, user, nil}
+      end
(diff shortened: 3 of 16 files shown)


skyqrose:
Just pushed a small fix and watched the sandbox deploy, and it did indeed force the page to refresh.
Console warning. Server has version pr-225-a086a68 but we have version pr-225-beb0d4a. Reloading.

js/socket.ts
```diff
@@ -0,0 +1,54 @@
...
+    console.warn("socket error");
+  });
+
+  socket.onOpen(() => {
+    socket
+      .channel("metadata")
+      .join()
+      .receive("ok", (rawResponse: unknown) => {
+        const { authenticated, server_release: serverRelease } =
+          MetadataJoinResponse.parse(rawResponse);
```
mathcolo, line 25: Would we prefer to catch zod parse errors here and log them to Sentry, but not throw a whole-page error and so on?

js/socket.ts
```diff
@@ -0,0 +1,54 @@
...
+        if (serverRelease !== clientRelease) {
+          console.warn(
+            `Server has version ${serverRelease} but we have version ${clientRelease}. Reloading.`,
+          );
+          reload();
+        }
+        if (!authenticated) {
+          console.warn("Authentication failed. Reloading.");
+          reload();
```
mathcolo, line 34: This makes me nervous but I don't have a good reason for that. we're absolutely positive that websocket connections get the same authentication treatment, _always_ on the server side?

lib/orbit_web/channels/user_socket.ex
```diff
@@ -0,0 +1,90 @@
...
+  require Logger
+
+  channel "metadata", OrbitWeb.MetadataChannel
+
+  @doc """
+  fields that are added to socket.assigns
+  """
+  @type metadata :: %{
+          server_release: String.t(),
+          client_release: String.t() | nil,
```
mathcolo, line 28: `client_release` could be nil? is that for running locally?

lib/orbit_web/channels/user_socket.ex
```diff
@@ -0,0 +1,90 @@
...
+
+        {:ok, authed_socket} ->
+          user = Guardian.Phoenix.Socket.current_resource(authed_socket)
+          {authed_socket, user, nil}
+      end
+
+    metadata = %{
+      server_release: server_release,
+      client_release: client_release,
+      release_matches: client_release == server_release,
```
mathcolo, line 57: should this only be if `client_release` is not `nil`?

lib/orbit_web/channels/user_socket.ex
```diff
@@ -0,0 +1,90 @@
...
+    server_release = Application.fetch_env!(:orbit, :release)
+    client_release = params["release"]
+
+    {socket, user, authentication_error} =
+      case Guardian.Phoenix.Socket.authenticate(socket, OrbitWeb.Auth.Guardian, params["token"]) do
+        {:error, :no_token} ->
+          {socket, nil, :no_token}
+
+        {:error, _e} ->
+          {socket, nil, :invalid_token}
```
mathcolo, line 47: Just to make sure I understand, this is sent on the already established socket yes?

js/socket.ts
skyqrose: This is already in an async handler, so if it fails it won't crash the whole app (and I just tested that). I'm not sure if that's a normal JS/browser thing or if it's because Phoenix has its own try/catch in `receive`.

Unfortunately, I don't think it gets sent to sentry. I'll look into that.

js/socket.ts
skyqrose: Yeah, this is a bit risky. During testing, where I forced auth to fail, this turned into an infinite reload loop, which made the screen flash repeatedly and obnoxiously.

If there aren't any bugs in our authentication, then this is fine, but this is a particularly obnoxious result if there are any bugs, and auth problems are very precedented.

lib/orbit_web/channels/user_socket.ex
skyqrose: Locally, the client release is always `local` (and reloading is handled by phoenix instead). https://github.com/mbta/orbit/blob/5c972eafb88ce811f585bd58c953731a2702fba7/config/dev.exs#L8

This is if you join the socket without passing in the client release. That doesn't have to be allowed, it's not part of any expected failure modes, but I was trying to make it so there were no application-defined rejections, so that any socket rejections were definitely because of network problems.

lib/orbit_web/channels/user_socket.ex
skyqrose: I think that's implied by `server_release` not being `nil`, but I could explicitly check that, too.

lib/orbit_web/channels/user_socket.ex
skyqrose: It's logged and put into `socket.assigns`, but the frontend wouldn't make any decisions based on it, so I didn't end up adding it to the data sent by `metadata_channel`.

This is a good segue to a thing I was wondering about: There's a bunch of metadata fields that are calculated and logged here but not used for anything else:
- `authenticated` is sent to the frontend and is the only one that matters.
- `server_release` is sent to the frontend, but could be recalculated from the application env in `metadata_channel` instead of put in `assigns`.
- `user_id` is technically used in `id()`, but we don't use that function so I could remove that use.
- `client_release`/`release_matches` are used on the frontend, but the client already knows it, so it isn't sent.
- `email` is never used aside from the log, but is likely to be some day.
- `authentication_error` is never used aside from the log.

Should I remove any of the fields from the log and/or `assigns`?

js/socket.ts
mathcolo: To clarify, is it even possible that the cookie would be valid (what do we call that, the Guardian cookie?) but the token wouldn't be? Wondering if it could be useful to only redirect if the socket _and an HTTP call_ fail?

This feels like the Glides hacks again :( Perhaps it's just unavoidable for auto refresh?

lib/orbit_web/channels/user_socket.ex
mathcolo: Ah. That's fine, then

js/socket.ts
skyqrose: No it's not possible for the cookie to be valid and the token not to be: if you refresh, you'll get a new token based on the cookie. Well, not possible unless there are bugs, but auth is complex and bugs are possible.

You mean that if the token/socket auth failed, instead of reloading immediately, we fire off a HTTP call (which we expect to fail), and refresh. But if that request somehow succeeds, we direct to the login page? This is getting increasingly complex...

Another approach could be to store in localstorage that you just refreshed so that we can avoid refreshing more than once in a row? But again, complexity...

skyqrose: I think on the whole, this approach is less hacky than glides, because we get told by the server that we're not authenticated, instead of inferring it from which requests fail.

skyqrose:
I got the frontend tests working. (It's all in a new commit, no changes to the previous code.)

I still want to do one more manual test either locally or on sandbox before it's merged.

js/app.tsx
```diff
@@ -2,8 +2,11 @@
 import "./telemetry";
 import "./phoenix.js";
 import { App } from "./components/app.js";
+import { initSocket } from "./socket";
 import { createRoot } from "react-dom/client";
 
+initSocket();
```
mathcolo, line 8: Just checking, it's okay that this runs with no DOM set up yet?

js/socket.ts
mathcolo: Gotcha, yeah, I admit the bandaids for problems we haven't actually seen are not helpful. This should be fine

js/socket.ts
```diff
@@ -0,0 +1,54 @@
...
+        }
+        // channel successfully joined
+      })
+      .receive("timeout", (response: unknown) => {
+        console.warn("channel timeout");
+        console.warn(response);
+      })
+      .receive("error", (response: unknown) => {
+        console.warn("channel error");
+        console.warn(response);
```
mathcolo, line 44: Could this be `console.error` instead?

mathcolo (approved):

js/socket.ts
skyqrose: Oh, also, this specific bug we're afraid of would only be to a problem with Guardian authentication, which has been stable so far. Most auth problems come from interacting with the third party systems (keycloak+EntraID), which would cause auth to completely fail, not partially fail.

js/socket.ts
skyqrose: Ah, good catch. I was thinking of `warn` as expected problems and `error` as unexpected problems that indicate a bug, and this would only happen if the server rejects the channel join, which it shouldn't do.

skyqrose:
Deploying and re-deploying to sandbox worked as expected.

PR #228, opened 2025-01-16
feat: require override for no cert

mathcolo:
Asana Task: [🐞 Orbit must require override for no cert](https://app.asana.com/0/1206105669438487/1209151691501262/f)

⚠️ Due to squash merge of the previous PR into main: diff is good, list of commits is not
⚠️ There's unit-like tests in here, although I may still want an integration-like test. TBD.

image

- Add a `MissingCertification`, a `Certification` with `expires` omitted
- Factor out `Box`; create `MissingBox` and `ExpiryBox` with it
- Submit missing certifications to server
- Include "missing" in CSV export if no `expires`

Checklist

- Tests:
  - `(x)` Has tests
  - `( )` Doesn't need tests
- Product/Design sign off:
  - `( )` No review needed

### DIFF
--- a/js/components/operatorSignIn/attestation.tsx
+++ b/js/components/operatorSignIn/attestation.tsx
@@ -1,6 +1,6 @@
 import { useNow } from "../../dateTime";
 import { lookupDisplayName } from "../../hooks/useEmployees";
-import { Certification, getExpired } from "../../models/certification";
+import { Certification, filterExpired } from "../../models/certification";
 import { Employee } from "../../models/employee";
 import { className } from "../../util/dom";
 import { removeLeadingZero } from "../../util/string";
@@ -32,7 +32,7 @@ export const Attestation = ({
   const [bypass, setBypass] = useState<boolean>(false);
 
   const name = lookupDisplayName(badge, employees);
-  const expireds = getExpired(certifications, now);
+  const expireds = filterExpired(certifications, now);
 
   return (
     <div className="text-sm">
--- a/js/components/operatorSignIn/attestation.tsx
+++ b/js/components/operatorSignIn/attestation.tsx
@@ -1,6 +1,10 @@
 import { useNow } from "../../dateTime";
 import { lookupDisplayName } from "../../hooks/useEmployees";
-import { Certification, filterExpired } from "../../models/certification";
+import {
+  Certification,
+  filterExpired,
+  getMissing,
+} from "../../models/certification";
 import { Employee } from "../../models/employee";
 import { className } from "../../util/dom";
 import { removeLeadingZero } from "../../util/string";
@@ -32,6 +36,7 @@ export const Attestation = ({
   const [bypass, setBypass] = useState<boolean>(false);
 
   const name = lookupDisplayName(badge, employees);
+  const missing = getMissing(certifications, "blue");
   const expireds = filterExpired(certifications, now);
 
   return (
@@ -42,7 +47,7 @@ export const Attestation = ({
         ignoreExpired={bypass}
         now={now}
       />
-      {expireds.length === 0 || bypass ?
+      {(expireds.length === 0 && missing.length === 0) || bypass ?
         <>
           <SignInText />
           <form
@@ -91,6 +96,7 @@ export const Attestation = ({
           </ol>
           <Bypass
             expireds={expireds}
+            missing={missing}
             displayName={name}
             onContinue={function (): void {
               setBypass(true);
--- a/js/models/certification.ts
+++ b/js/models/certification.ts
@@ -1,15 +1,16 @@
+import { HeavyRailLine } from "../types";
 import { DateTime } from "luxon";
 import { z } from "zod";
 
 export type Certification = {
   type: string;
-  railLine: string;
+  railLine: HeavyRailLine;
   expires: DateTime;
 };
 
 export const CertificationData = z.object({
   type: z.string(),
-  rail_line: z.string(),
+  rail_line: HeavyRailLine,
   expires: z.string(),
 });
 export type CertificationData = z.infer<typeof CertificationData>;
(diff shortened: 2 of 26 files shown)


lib/orbit_web/controllers/sign_in_controller.ex
```diff
@@ -65,9 +65,8 @@ defmodule OrbitWeb.SignInController do
   @spec validate_override(map()) :: boolean()
   defp validate_override(override) do
     Enum.all?(override, fn cert ->
-      Util.Map.has_exactly?(cert, MapSet.new(["type", "expires", "rail_line"])) and
```
skyqrose, line 68: This is the only use of `has_exactly?`. We should probably remove it.

lib/orbit_web/controllers/sign_in_export_controller.ex
```diff
@@ -38,7 +38,7 @@ defmodule OrbitWeb.SignInExportController do
         ""
 
       cert ->
-        cert["expires"]
+        cert["expires"] || "missing"
```
skyqrose, line 41: So a override for a missing cert is stored as an override with `"expires" => nil`? That seems very implicit.

Where it described how this is stored in the database? `operator_sign_in.ex` just has `override: [map()] | nil`. I'm hoping for a struct, typespec, or even just a comment that describes what fields there are in that map, and how overriding missing certifications are represented.

js/components/operatorSignIn/attestation.tsx
```diff
@@ -32,6 +36,7 @@ export const Attestation = ({
   const [bypass, setBypass] = useState<boolean>(false);
 
   const name = lookupDisplayName(badge, employees);
+  const missing = getMissing(certifications, "blue");
```
skyqrose, line 39: Optional refactoring opportunity:

I was suprised to see 4 new uses of the new `getMissing` function. I guess I was expecting there to only be one or two. Generally when interesting logic like this is called from multiple places I worry about it getting out of sync in multiple places. Like, the business logic should be done once at the highest level possible, and then all the components below that should just be about displaying that data.

- Attestation takes `certifications` and recalculates the missing certs, but it also takes `onSubmit()`, which the caller gives in a reference to its version of `missing` via a closure. Should both of those be the same version of `missing`?
- In `expiry.tsx`, `Bypass` takes in `missing` as a param, but another component in the same file, `CertificationBoxes` takes the full list of certifications and recalculates the list of missing by itself.

But it's not a huge deal in this case:
- It's a pure function, there's no worry about double triggering effects.
- The logic is abstracted and DRYed into `certification.ts`.
- It's paralleling the list of expired certifications which already follows this pattern, so this isn't a new issue in this PR.
- In at least one case (CertificateBoxes), the subcomponent does logic that's different from the caller (showing warnings for certs that are about to expire but shouldn't block the sign in).

I guess if you decided to refactor this, `operatorSignInModal` would do one set of calculations on `certifications` to determine `missing`, `expired`, about to expire certs, and a boolean for whether to block sign in, and then do prop drilling to get all of those to the right places?

js/test/components/operatorSignIn/attestation.test.tsx
```diff
@@ -4,9 +4,21 @@ import { Employee } from "../../../models/employee";
...
+    expires: DateTime.fromISO("2030-01-01"),
+    railLine: "blue",
+    type: "rail",
+  },
+  {
+    expires: DateTime.fromISO("2030-01-01"),
+    railLine: "none",
+    type: "right_of_way",
+  },
+];
```
skyqrose, line 21: Hm? This isn't `_NONE` anymore. What's happening?

js/models/certification.ts
```diff
@@ -3,12 +3,21 @@ import { HeavyRailLine } from "../types";
...
 
+const CERT_TYPES = ["right_of_way", "rail"];
+
 export type Certification = {
   type: string;
   railLine: HeavyRailLine;
   expires: DateTime;
 };
 
+export type MissingCertification = Omit<Certification, "expires">;
```
skyqrose, line 14: This type is a bit weird in that unlike Certification, it doesn't describe data that comes from the server, it's data that the frontend creates during its calculations and then sends to the backend.

Is that worth calling out in a `/** doc comment */`?

skyqrose (commented):
Sorry this took so long, it took me a few tries to figure out what felt off about `getMissing`.

js/components/operatorSignIn/attestation.tsx
mathcolo: Yeah, I'm definitely appreciative of this. It's not the best long-term structure at all. How would you feel about prop drilling a, say, `CertificateStatus` that contains `{expired: [certs], missing: [certs]}`?

js/test/components/operatorSignIn/attestation.test.tsx
mathcolo: Ah, yeah, the "NONE" state was a kind of catch-all that before worked for testing. But with certifications missing, it actually became an error state, so it needed some. Changed the name 👍

lib/orbit_web/controllers/sign_in_export_controller.ex
mathcolo: Agreed, this is opaque. I added a comment to the [typespec](https://github.com/mbta/orbit/pull/228/files#diff-11f51221667403a857be3385878d97c5f25283655ef63a5a77c4a0cd88fafc6dR23); does that feel satisfying?

js/components/operatorSignIn/attestation.tsx
skyqrose: Sure. You may also want the "about to expire" certs in that object so that CertificateBoxes doesn't have to take all the certs and do its own calculations. I suggested a `requireOverride` boolean but that's less important cuz it's easy to calculate from CertificateStatus.

Perhaps this should be a separate followup refactor PR rather than blocking this PR.

skyqrose (approved):
👍 See the outstanding comment, you could merge this as is and do that refactor as a followup to avoid blocking this PR and make that refactor a little easier to review.

js/components/operatorSignIn/attestation.tsx
mathcolo: this is nice! https://github.com/mbta/orbit/pull/244

PR #244, opened 2025-01-28
refactor: encapsulate expired and missing certs into CertificationStatus

mathcolo:
Asana Task: Follow up on [🐞 Orbit must require override for no cert](https://app.asana.com/0/1206105669438487/1209151691501262/f)

Refactor as per https://github.com/mbta/orbit/pull/228#discussion_r1932285459.

- Create a `CertificationStatus`
- Pass it down through child component hierarchy
- Update tests

Checklist

- Tests:
  - `(x)` Has tests
  - `( )` Doesn't need tests
- Product/Design sign off:
  - `(x)` No review needed


js/components/operatorSignIn/operatorSignInModal.tsx
```diff
@@ -204,15 +209,14 @@ const OperatorSignInModalContent = ({
...
-              getMissing(certifications.result, "blue"),
+              certificationStatus,
               setComplete,
               setLoading,
               onComplete,
             );
           }}
           employees={employees.result}
-          certifications={certifications.result}
+          certificationStatus={getStatus(certifications.result, now, "blue")}
```
skyqrose, line 219: Why recalculate this instead of using the variable `certificationStatus` from earlier?

js/components/operatorSignIn/operatorSignInModal.tsx
```diff
@@ -37,17 +35,20 @@ enum CompleteState {
 const submit = (
   badgeEntry: BadgeEntry,
   radio: string,
-  expired: Certification[],
-  missing: MissingCertification[],
+  certificationStatus: CertificationStatus | null,
```
skyqrose, line 38: This feels a bit weird. If certificationStatus is null (cuz it's loading), then you shouldn't submit, that makes sense. What do you think about making it the caller's responsibility to not call `submit()` in that case instead of taking the null and doing nothing with it? That might make it more likely that the caller remembers to have a disabled button or something.

js/components/operatorSignIn/expiry.tsx
```diff
@@ -136,17 +129,24 @@ export const CertificateBoxes = ({
...
           mode="error"
-          title={expired.length === 1 ? "Expired card" : "Expired cards"}
+          title={
+            certificationStatus.expired.length === 1 ?
+              "Expired card"
+            : "Expired cards"
+          }
           operatorName={displayName}
-          certifications={expired}
+          certifications={certificationStatus.expired}
```
skyqrose, line 142: Just a tiny style thought: What do you think about doing `const {expired, missing} = certificationStatus;` above so you can keep referring to these as `expired`.

skyqrose (approved):
All comments optional, no rereview needed.

PR #282, opened 2025-04-07
feat: jcc set up polling gen server

cho-jos:
Asana Task: [Set up the polling GenServer for vehicle positions and trip updates](https://app.asana.com/0/1200273269966439/1209646113903406)

Checklist

- Tests:
  - `(X)` Has tests
  - `( )` Doesn't need tests
- Product/Design sign off:
  - `(X)` No review needed

### DIFF
--- a/config/dev.exs
+++ b/config/dev.exs
@@ -47,7 +47,8 @@ config :orbit, OrbitWeb.Endpoint,
 
 config :orbit, Orbit.S3,
   folders: [
-    glides_global: "local-s3-stub://glides/global/"
+    glides_global: "local-s3-stub://glides/global/",
+    rtr_public: "s3-anonymous://mbta-gtfs-s3/"
   ]
 
 # CSV import jobs
@@ -88,6 +89,9 @@ config :ueberauth_oidcc,
       client_secret: "dev-secret"
     ]
   ]
+
+config :orbit,
+  log_polls?: false
 
 if File.exists?(Path.expand("dev.secret.exs", __DIR__)) do
   import_config "dev.secret.exs"
--- a/config/runtime.exs
+++ b/config/runtime.exs
@@ -41,7 +41,7 @@ if config_env() == :prod do
   config :orbit, Orbit.S3,
     folders: [
       glides_global: "s3://#{System.fetch_env!("GLIDES_BUCKET")}/global",
-      rtr_public: System.get_env("S3_RTR_PUBLIC")
+      rtr_public: System.fetch_env!("S3_RTR_PUBLIC")
     ]
 
   # CSV import jobs
--- a/config/runtime.exs
+++ b/config/runtime.exs
@@ -41,7 +41,7 @@ if config_env() == :prod do
   config :orbit, Orbit.S3,
     folders: [
       glides_global: "s3://#{System.fetch_env!("GLIDES_BUCKET")}/global",
-      rtr_public: "s3-anonymous://mbta-gtfs-s3/"
+      rtr_public: System.get_env("S3_RTR_PUBLIC")
     ]
 
   # CSV import jobs
--- a/config/runtime.exs
+++ b/config/runtime.exs
@@ -40,7 +40,8 @@ if config_env() == :prod do
   # S3
   config :orbit, Orbit.S3,
     folders: [
-      glides_global: "s3://#{System.fetch_env!("GLIDES_BUCKET")}/global"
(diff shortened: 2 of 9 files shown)


config/runtime.exs
```diff
@@ -40,7 +40,8 @@ if config_env() == :prod do
   # S3
   config :orbit, Orbit.S3,
     folders: [
-      glides_global: "s3://#{System.fetch_env!("GLIDES_BUCKET")}/global"
+      glides_global: "s3://#{System.fetch_env!("GLIDES_BUCKET")}/global",
+      rtr_public: System.get_env("S3_RTR_PUBLIC")
```
mathcolo, line 44: question: does this also need to be an S3 URI like `s3://...`?

config/test.exs
```diff
@@ -26,7 +26,8 @@ config :orbit, Oban, testing: :inline
 
 config :orbit, Orbit.S3,
   folders: [
-    glides_global: "local-s3-stub://glides/global/"
+    glides_global: "local-s3-stub://glides/global/",
+    rtr_public: "s3-anonymous://mbta-gtfs-s3/"
```
mathcolo, line 30: issue: tests probably shouldn't try to connect to S3 directly

config/dev.exs
```diff
@@ -47,7 +47,8 @@ config :orbit, OrbitWeb.Endpoint,
 
 config :orbit, Orbit.S3,
   folders: [
-    glides_global: "local-s3-stub://glides/global/"
+    glides_global: "local-s3-stub://glides/global/",
+    rtr_public: "s3-anonymous://mbta-gtfs-s3/"
```
mathcolo, line 51: issue: should this also be `local-s3-stub`?

config/runtime.exs
mathcolo: suggestion: [fetch_env!/1](https://hexdocs.pm/elixir/1.18.3/System.html#fetch_env!/1) raises instead of returning nil, which we may want here (like the line above)

cho-jos: > question: does this also need to be an S3 URI like `s3://...`?

I reverse-engineered this from how Glides handles it, so the variable definition in the orbit [terraform_module](https://github.com/mbta/terraform_modules/blob/main/terraform-aws-mbta-orbit/main.tf#L44) formats the link in S3 URI format already

> suggestion: [fetch_env!/1](https://hexdocs.pm/elixir/1.18.3/System.html#fetch_env!/1) raises instead of returning nil, which we may want here (like the line above)

good suggestion, thanks! like the above, this is just me reverse-engineering from Glides (a little too exactly in this case)

config/test.exs
cho-jos: gotcha -- like everything above this is also reverse-engineering what Glides does too closely

config/dev.exs
cho-jos: I followed what Glides does in its [dev.exs](https://github.com/mbta/glides/blob/main/config/dev.exs#L94) counterpart -- I had made the assumption this connection is in place so that local glides can test making s3 connections

cho-jos:
not sure why the docker container fails to built (I tried giving it `S3_RTR_PUBLIC`) 🤔

config/dev.exs
mathcolo: oh! makes sense 👍

mathcolo:
[This MBTA RFC standardizing on colima](https://github.com/mbta/technology-docs/blob/main/rfcs/accepted/0010-docker-desktop-replacement.md) should help you try this locally, let me know if you get stuck 😃 my initial hunch is that it's configuration related, but tough to tell with GitHub Actions output not showing the Phoenix logging.

mathcolo:
Should be good to approve once https://github.com/mbta/devops/pull/2749 is in?

mathcolo (approved):

test/realtime/polling_server_test.exs
```diff
@@ -0,0 +1,77 @@
...
+
+                 # update this later to build(:vehicle_position)
+                 entities: []
+               }
+           end
+         }},
+        id: :vehicle_positions_test
+      )
+
+    # PollingServer.subscribe(self(), :vehicle_positions_test)
```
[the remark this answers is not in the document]
cho-jos: it can definitely be removed. I think that made it in there because (again 😅) i was reverse engineering from Glides and wasn't sure if at this stage the initial polling_server test for orbit needed to subscribe to the `:vehicle_positions_test` -- but that can be added in subsequent orbit work

PR #285, opened 2025-04-16
feat: jcc create rtr data structs

cho-jos:
Asana Task: [Set up VehiclePosition, TripUpdate, and Prediction structs](https://app.asana.com/1/15492006741476/project/1200273269966439/task/1209660480370358?focus=true)

Description goes here.
Include screenshots if relevant.

Checklist

- Tests:
  - `(X)` Has tests
  - `( )` Doesn't need tests
- Product/Design sign off:
  - `(X)` No review needed


lib/realtime/data/stations.ex
```diff
@@ -0,0 +1,54 @@
+defmodule Realtime.Data.Stations do
+  @spec platforms_to_stations :: %{required(String.t()) => String.t()}
+  def platforms_to_stations do
+    %{
+      # southbound platform followed by northbound platform
+      "70061" => "place-alfcl",
```
mathcolo, line 6: question: is 70062 the other platform at Alewife?

lib/realtime/data/stations.ex
```diff
@@ -0,0 +1,54 @@
...
+      "70096" => "place-jfk",
+      "70097" => "place-nqncy",
+      "70098" => "place-nqncy",
+      "70099" => "place-wlsta",
+      "70100" => "place-wlsta",
+      "70101" => "place-qnctr",
+      "70102" => "place-qnctr",
+      "70103" => "place-qamnl",
+      "70104" => "place-qamnl",
+      "70105" => "place-brntn"
```
mathcolo, line 51: question: Likewise, is there another Braintree platform?

lib/realtime/data/trip_update.ex
```diff
@@ -0,0 +1,70 @@
...
+
+  @enforce_keys [
+    :label,
+    :route_id
+  ]
+  defstruct [
+    :label,
+    :route_id,
+    direction: nil,
+    stop_time_updates: []
```
mathcolo, line 21: question: Are the latter two of these per-initialized? what's the difference between specifying `: nil` and specifying nothing?

lib/realtime/data/trip_update.ex
```diff
@@ -0,0 +1,70 @@
+defmodule Realtime.Data.TripUpdate do
+  @moduledoc """
+  The subset of the TripUpdate data in the enhanced-json feed that we use.
+  """
+
+  @type t :: %__MODULE__{
```
mathcolo, line 6: question: I noticed trip_id is missing here, do we need it, or is this struct typically nested inside another struct that has that information?

lib/realtime/data/vehicle_position.ex
```diff
@@ -0,0 +1,34 @@
+defmodule Realtime.Data.VehiclePosition do
+  @moduledoc """
+  The subset of the VehiclePosition data in the enhanced_json feed that we use.
+  """
+  use Ecto.Schema
+
+  @type t :: %__MODULE__{
+          route_id: Realtime.Data.route_id() | nil,
```
mathcolo, line 8: issue: Maybe we should require a `route_id` and therefore remove the `| nil`?

lib/realtime/data/vehicle_position.ex
```diff
@@ -0,0 +1,34 @@
...
+          direction: integer(),
+          label: String.t(),
+          position: Util.Position.t() | nil,
```
mathcolo, line 11: question: Same with the `| nil` here. If Glides has these, I'd be curious to find out why?

lib/realtime/data/vehicle_position.ex
```diff
@@ -0,0 +1,34 @@
...
+          heading: float() | nil,
+          station_id: String.t() | nil,
+          current_status: :INCOMING_AT | :STOPPED_AT | :IN_TRANSIT_TO,
+          timestamp: DateTime.t() | nil,
```
mathcolo, line 15: question: Same with this `| nil` here?

lib/realtime/data/vehicle_position.ex
```diff
@@ -0,0 +1,34 @@
...
+          # GTFS-RT id, e.g. "R-54826B16".
+          vehicle_id: String.t() | nil,
```
mathcolo, line 17: praise: This `| nil` makes more sense, as we may still be able to show the vehicle on the ladder even if it has no GTFS-RT/OCS id.

lib/realtime/data/vehicle_position.ex
```diff
@@ -0,0 +1,34 @@
...
+          trip: nil
+        }
+
+  schema "vehicle_positions" do
+    field(:route_id, Util.AtomType)
+    field(:direction, :integer)
+    field(:label, :string, default: "")
```
mathcolo, line 24: question: Not sure we want to default to empty string, should we filter out vehicle positions with no label?

lib/realtime/data/vehicle_position.ex
```diff
@@ -0,0 +1,34 @@
...
+          direction: integer(),
+          label: String.t(),
+          position: Util.Position.t() | nil,
+          heading: float() | nil,
+          station_id: String.t() | nil,
+          current_status: :INCOMING_AT | :STOPPED_AT | :IN_TRANSIT_TO,
+          timestamp: DateTime.t() | nil,
+          # GTFS-RT id, e.g. "R-54826B16".
+          vehicle_id: String.t() | nil,
+          trip: nil
```
mathcolo, line 18: note: `trip: nil` in the typespec doesn't make sense in the long term, as it means "trip is present, but it always nil". In Glides this is how the server-side trip match is sent to the client (it's a `virtual` field, so it isn't persisted), but it does make sense that we'll ultimately be putting OCS trip data here. So we can leave it 👍

lib/realtime/rtr.ex
```diff
@@ -3,14 +3,93 @@ defmodule Realtime.RTR do
...
+
+  def parse_vehicle_positions(filedata) do
+    json = Jason.decode!(filedata)
+
+    %{
+      timestamp: json["header"]["timestamp"],
+      entities:
+        json["entity"]
+        |> Enum.map(&parse_vehicle_position/1)
+        |> Enum.filter(fn vp -> vp.route_id != nil end)
```
mathcolo, line 23: note: Ah, here is where `route_id` cannot be `nil`. So that can be reflected in the spec 👍

lib/realtime/rtr.ex
```diff
@@ -3,14 +3,93 @@ defmodule Realtime.RTR do
...
+  @spec parse_vehicle_position(map()) :: VehiclePosition.t()
+  def parse_vehicle_position(entity_json) do
+    vehicle = entity_json["vehicle"]
+
+    %VehiclePosition{
+      route_id: Realtime.Data.route_id_from_string(vehicle["trip"]["route_id"]),
+      direction: vehicle["trip"]["direction_id"],
+      label: vehicle["vehicle"]["label"],
+      position: parse_position(vehicle["position"]),
+      heading: vehicle["position"]["bearing"],
```
mathcolo, line 36: issue: Heading and bearing are different, does that need to be converted?

lib/realtime/rtr.ex
```diff
@@ -3,14 +3,93 @@ defmodule Realtime.RTR do
...
+  defp parse_stop_time_update(stu_json) do
+    %Realtime.Data.TripUpdate.StopTimeUpdate{
+      station_id: Realtime.Data.Stations.platforms_to_stations()[stu_json["stop_id"]],
+      predicted_arrival_time: stu_json["arrival"] && stu_json["arrival"]["time"],
+      predicted_departure_time: stu_json["departure"] && stu_json["departure"]["time"]
     }
   end
+
+  @spec parse_position(map() | nil) :: Util.Position.t() | nil
+  defp parse_position(%{"latitude" => latitude, "logitude" => longitude}) do
```
mathcolo, line 85: issue: Typo (`lo_gitude`) here, which may mean this needs a test if it didn't fail 😃

lib/realtime/data/stations.ex
cho-jos: same thing as above with Alewife -- I can only find one stop_id associated with Braintree

lib/realtime/rtr.ex
cho-jos: This is actually just another thing I pulled from [Glides](https://github.com/mbta/glides/blob/main/lib/realtime/rtr.ex#L37) so perhaps I'm just misunderstanding here

mathcolo: Clarified 😃 https://mbta.slack.com/archives/G8Y8HEY2D/p1744991523015939

lib/realtime/data/trip_update.ex
mathcolo: OK, I thought about this more. So we need the trip ID to connect to OCS trips. But it remains true that we have a `trip` key in `VehiclePosition`, so TripUpdates (predictions) can probably go without it. (We probably want to connect positions to predictions by consist, not trip)

lib/realtime/data/stations.ex
mathcolo: hmm. https://mbta.slack.com/archives/G8Y8HEY2D/p1744992935244699

lib/realtime/rtr.ex
cho-jos: good catch -- added a test to explicitly checking for correctly parsing the structs

lib/realtime/data/trip_update.ex
cho-jos: I took this to mean that `label` and `route_id` are non-negotiables here and thus are covered in `@enforce_keys` above. On the other hand, direction and stop_time_updates may fluctuate based on realtime changes (updates from the rtr file) so those are initialized to some value that (hopefully) updates

lib/realtime/data/vehicle_position.ex
cho-jos: Sorry -- I may have lost the thread (literally) -- what was the original question?

mathcolo: No prob 😅 This was the original: https://github.com/mbta/orbit/pull/285#discussion_r2049386034

lib/realtime/data/trip_update.ex
cho-jos: > question: Are the latter two of these per-initialized? what's the difference between specifying : nil and specifying nothing?
[clarified](https://mbta.slack.com/archives/C06Q8TA1PPV/p1745345722363999?thread_ts=1745345361.896209&cid=C06Q8TA1PPV) that there is no difference. I'm comfortable keeping these since `direction` and `stop_time_updates` are the 2 non-required fields so just initialize them to something. Lmk thoughts

lib/realtime/data/vehicle_position.ex
cho-jos: doesn't seem to be strong historical reason to allow these fields to be nillable/not required. will update!

lib/realtime/data/vehicle_position.ex
```diff
@@ -0,0 +1,34 @@
+defmodule Realtime.Data.VehiclePosition do
+  @moduledoc """
+  The subset of the VehiclePosition data in the enhanced_json feed that we use.
+  """
+  use Ecto.Schema
+
+  @type t :: %__MODULE__{
+          route_id: Realtime.Data.route_id(),
+          direction: integer(),
+          label: String.t(),
```
mathcolo, line 10: It occurred to me that "label" for LR is a `-`-split list of the consist, but on HR it's just the lead car number. So label is fine to leave, but we'll still need a way to represent car numbers beyond the lead car in here. Maybe [a PostgreSQL array](https://www.postgresql.org/docs/current/arrays.html) would be okay here? What do you think?

cho-jos: gotcha, that makes sense. do you still think it's worth it to have `label` as its own separate field?

lib/realtime/rtr.ex
```diff
@@ -3,14 +3,99 @@ defmodule Realtime.RTR do
...
+    %Realtime.Data.TripUpdate.StopTimeUpdate{
+      station_id: Realtime.Data.Stations.platforms_to_stations()[stu_json["stop_id"]],
+      predicted_arrival_time: stu_json["arrival"] && stu_json["arrival"]["time"],
+      predicted_departure_time: stu_json["departure"] && stu_json["departure"]["time"]
+    }
+  end
+
+  @spec parse_cars([map()]) :: [String.t()]
+  defp parse_cars(carriage_details) do
+    carriage_details |> Enum.map(fn car -> car["label"] end)
```
mathcolo, line 87: **nit:** We generally try to avoid use of a single pipe operator in a statement; i.e. only 2 or more. So this could instead be written as this, also using the [capture operator](https://hexdocs.pm/elixir/1.18.3/anonymous-functions.html#the-capture-operator) for shorthand:
```
Enum.map(carriage_details, & &1["label"])
```

test/realtime/rtr_test.exs
```diff
@@ -0,0 +1,74 @@
...
+  describe "VehiclePositions decode from json" do
+    setup do
+      {:ok, vehicle_positions} = File.read(__DIR__ <> "/VehiclePositions_enhanced.json")
+      {:ok, vehicle_positions: RTR.parse_vehicle_positions(vehicle_positions)}
+    end
+
+    test "correctly parses single entity", context do
+      vp =
+        context[:vehicle_positions][:entities]
+        |> Enum.filter(fn vp -> vp.timestamp == ~U[2025-04-23 16:06:35Z] end)
```
mathcolo, line 15: **nit:** another place we may prefer avoiding `|>`

test/realtime/rtr_test.exs
```diff
@@ -0,0 +1,74 @@
...
+  end
+
+  describe "TripUpdates decode from json" do
+    setup do
+      {:ok, trip_updates} = File.read(__DIR__ <> "/TripUpdates_enhanced.json")
+      {:ok, trip_updates: RTR.parse_trip_updates(trip_updates)}
+    end
+
+    test "correctly parses single entity", context do
+      tu = context[:trip_updates][:entities] |> Enum.at(0)
```
mathcolo, line 52: **nit:** same here

mathcolo (approved):

lib/realtime/rtr.ex
cho-jos: dang, I actually now remember that point about using `|>` for >= 2 transformations. I'll update the code as such

On the point of the capture operator shorthand -- that is _very_ cool 🤯🤯🤯 . However, I almost want to refrain from using it here in favor of writing out the anon `fn` for readability/understandability. <-- thoughts?

mathcolo: Readability fine by me! 😃 🚀

PR #287, opened 2025-04-18
refactor: replace initSocket with a new SocketProvider

mathcolo:
Asana Task: [Set up the Elixir/Phoenix socket](https://app.asana.com/1/15492006741476/project/1206105669438487/task/1209646113903408)

Orbit currently has an initSocket that sets up the metadata socket in global scope. This moves it to React-land.

Note: This does not support gzipped message payloads. We can add that later if we want it.

Checklist

- Tests:
  - `(x)` Has tests
  - `( )` Doesn't need tests
- Product/Design sign off:
  - `( )` No review needed


mathcolo:
@skyqrose Joe and I went over this, but could you validate the approach?

skyqrose (commented):
This `canReachGlides` approach caused a ton of problems in Glides, so I tried a different approach for Orbit that can distinguish between network failure and auth problems. Here's how it worked before this PR:

- The socket connection always succeeds if and only if there's internet.
- In `js/socket.ts`, after connecting to the socket, it immediately joins a `"metadata"` channel. The response to that channel can be anything we want, so we _do_ indicate authentication failures (and also outdated versions), and trigger a reload.
- Conceptually, the socket is only fully initialized after the metadata channel is joined and gets a successful reply. (But nothing else uses the socket yet, so it's never returned, and there's no "fully initialized socket" value in the code.)

I think that approach was working well. Unless you have a good reason to go back to the Glides approach, I recommend continuing it.

It would still need to be translated into React. I think the way to do that would be for the `SocketProvider` to only provide the socket once it's fully initialized with the metadata channel (and therefore authorized). Before that point, it holds the partially-initialized socket in internal state only.

That would remove the need for `canReachOrbit`.

(and `js/socket.ts` will probably need to get deleted cuz all of its functionality would be moved into `SocketProvider`.)

mathcolo:
Ah, gotcha. Thanks for the clarifying that. This PR's changes were unintentionally erroneous; I forgot that the metadata socket's carrying authentication state is exactly _for_ this purpose.

js/contexts/socketContext.tsx
```diff
@@ -0,0 +1,66 @@
...
+        release: clientRelease,
+        token: getMetaContent("guardianToken"),
+      },
+    });
+    initialSocket.connect();
+
+    initialSocket.onError((err) => {
+      console.error(err);
+    });
+    const metadataChannel = initialSocket.channel("metadata");
```
skyqrose, line 37: Connecting the socket is asynchronous, so the socket might not be connected yet. Will this automatically wait until the socket is connected before making and joining the channel? Or do you have to put it in a `socket.onOpen()` callback?

js/contexts/socketContext.tsx
```diff
@@ -0,0 +1,66 @@
...
+    // Reload our session if the auth has expired
+    metadataChannel.on("auth_expired", () => {
```
skyqrose, line 39: Is this a phoenix thing or an orbit thing? Cuz the backend in `user_socket.ex` and `metadata_channel.ex` doesn't mention the string `"auth_expired"`. I think it successfully allows joining the metadata channel even when unauthenticated, and just returns  `%{authenticated: false}` (which you check below).

It might be vestigial and deleteable.

js/hooks/useChannel.ts
```diff
@@ -0,0 +1,90 @@
...
+ * WARNING: If you join a channel which is already joined, the existing subscription will be silently replaced, which is not what you want.
+ */
+export const useRawChannel = ({
+  socket,
+  actuallyConnect = true,
+  joinParams,
+  topic,
+}: {
+  socket: Socket | null;
+  actuallyConnect?: boolean;
```
skyqrose, line 22: I think YAGNI for `actuallyConnect`.

It has a few uses in Glides:
- Scenario mode
- Manage runs when looking at tomorrow (which doesn't have live data)
- The realtime map page (when you aren't showing some data on the map)
- The predictions channel, when the panel is closed.

You _might_ need it some day, but I don't think any of those situations will pop up, and it'll be easy enough to re-add if needed.

js/contexts/socketContext.tsx
mathcolo: Ooh, good question. In looking at the phoenix(js) source, channel creation doesn't cause any message send. And Channel#join, which uses Socket#push, is [buffered](https://github.com/phoenixframework/phoenix/blob/37c6085414daefdc88a4a303c876a616d425a4d7/assets/js/phoenix/socket.js#L607)!

js/contexts/socketContext.tsx
mathcolo: Yep! thanks 👍 Deleted.

skyqrose (approved):

PR #291, opened 2025-04-28
feat: Jcc ladder scaffold

cho-jos:
Asana Task: [Implement ladder scaffold](https://app.asana.com/1/15492006741476/project/1200273269966439/task/1209815362751264?focus=true)

Adds the ladder scaffold for the 3 sections of the red line ladder to `/ladder`:
- Alewife-Andrew
- JFK-Ashmont
- JFK-Braintree

Screenshot 2025-04-28 at 5 20 38 PM

Checklist

- Tests:
  - `(X)` Has tests
  - `( )` Doesn't need tests
- Product/Design sign off:
  - `(X)` No review needed


mix.lock
```diff
@@ -52,7 +52,7 @@
...
   "phoenix_template": {:hex, :phoenix_template, "1.0.4", "e2092c132f3b5e5b2d49c96695342eb36d0ed514c5b252a77048d5969330d639", [:mix], [{:phoenix_html, "~> 2.14.2 or ~> 3.0 or ~> 4.0", [hex: :phoenix_html, repo: "hexpm", optional: true]}], "hexpm", "2c0c81f0e5c6753faf5cca2f229c9709919aba34fab866d3bc05060c9c444206"},
   "plug": {:hex, :plug, "1.17.0", "a0832e7af4ae0f4819e0c08dd2e7482364937aea6a8a997a679f2cbb7e026b2e", [:mix], [{:mime, "~> 1.0 or ~> 2.0", [hex: :mime, repo: "hexpm", optional: false]}, {:plug_crypto, "~> 1.1.1 or ~> 1.2 or ~> 2.0", [hex: :plug_crypto, repo: "hexpm", optional: false]}, {:telemetry, "~> 0.4.3 or ~> 1.0", [hex: :telemetry, repo: "hexpm", optional: false]}], "hexpm", "f6692046652a69a00a5a21d0b7e11fcf401064839d59d6b8787f23af55b1e6bc"},
-  "plug_crypto": {:hex, :plug_crypto, "2.1.1", "19bda8184399cb24afa10be734f84a16ea0a2bc65054e23a62bb10f06bc89491", [], [], "hexpm", "6470bce6ffe41c8bd497612ffde1a7e4af67f36a15eea5f921af71cf3e11247c"},
+  "plug_crypto": {:hex, :plug_crypto, "2.1.1", "19bda8184399cb24afa10be734f84a16ea0a2bc65054e23a62bb10f06bc89491", [:mix], [], "hexpm", "6470bce6ffe41c8bd497612ffde1a7e4af67f36a15eea5f921af71cf3e11247c"},
```
cho-jos, line 55: note: I did not manually add this change, unsure what added it

js/components/ladderPage/ladder.tsx
```diff
@@ -1,13 +1,65 @@
...
 import { ReactElement } from "react";
 
 export const Ladder = (): ReactElement => {
   const vehiclePositions = useVehiclePositions();
-
   return (
     <>
-      Ladder Page.
+      <div className="flex justify-center">
+        <StationLists stations={alewifeAndrewStationsArray} />
```
mathcolo, line 16: **nit:** `StationLists` could instead be `StationList`, as each is one list (singular)?

js/data/stations.ts
```diff
@@ -0,0 +1,39 @@
+import { Station } from "../models/station";
+
+const buildStation = (id: string, name: string, spacing: number): Station => ({
+  id,
+  name,
+  spacing,
+});
+
+export const alewifeAndrewStationsArray: Station[] = [
```
mathcolo, line 9: suggestion: `alewifeAndrewStationsArray` is kind of weird name, could this instead be something more akin to this?
```
export const StationSets = {
  AlewifeAndrew: [
     buildStation(...),
     buildStation(...),
  ],
  JFKAshmont: [
     ...
  ],
  JFKBraintree: [
     ...
    ],
};
```

And then referred to as `StationSets.AlewifeAndrew`?

js/components/ladderPage/ladder.tsx
```diff
@@ -1,13 +1,65 @@
...
+      </div>
       <div>Vehicle positions: {vehiclePositions}</div>
     </>
   );
 };
+
+const StationLists = ({ stations }: { stations: Station[] }): ReactElement => {
+  return (
+    <div className="relative mt-20 mb-20">
+      <ul className="relative mx-36 w-32 border-x-[6px] border-solid border-gray-300">
```
mathcolo, line 28: **question:** do you need [relative](https://developer.mozilla.org/en-US/docs/Web/CSS/position#relative) here? That positioning usually lets you nudge something by a top/bottom/left/right, but I don't see any of those here

js/components/ladderPage/ladder.tsx
```diff
@@ -1,13 +1,65 @@
...
+        <li className="pt-20" />
+
+        {/* northbound arrow on top right */}
+        <div className="absolute right-[-18px] top-0 w-0 h-0 border-l-[15px] border-l-white border-r-[15px] border-r-white border-b-[25px] border-b-gray-300"></div>
```
mathcolo, line 32: **suggestion:** these two divs can be self-closing

mix.lock
mathcolo: **comment:** I've seen this before, too; it's no problem!

js/components/ladderPage/ladder.tsx
```diff
@@ -1,13 +1,65 @@
...
+              {/* station dots on left and right */}
+              <div className="absolute bg-white mt-0.5 left-[-13px] size-5 rounded-full border-4 border-gray-300" />
+              <div className="absolute bg-white mt-0.5 right-[-13px] size-5 rounded-full border-4 border-gray-300" />
+              <div className="text-center mx-auto text-wrap w-24">
+                {station.name}
+              </div>
+            </li>
+          );
+        })}
+        <li className="pb-20" />
```
mathcolo, line 61: **question:** is this here to extend the line beyond the last station dot? although it may be visually correct, from a "what does this HTML _mean_" perspective (i.e. [Semantic HTML](https://en.wikipedia.org/wiki/Semantic_HTML)) it looks like it's another station. And a screen reader or other accessibility software may think the same. is there a way to do this without making another list item?

js/components/ladderPage/ladder.tsx
mathcolo, line 28: **question:** if the div has one child element, the unordered list, maybe the parent div should be the unordered list itself? And you could remove the parent div?

cho-jos: yes I do need the `relative` here otherwise the direction arrows and station dots are not correctly aligned with the `<ul>` border:
Screenshot 2025-04-28 at 1 58 23 PM

as far as I understand it, this is because I used `absolute` for the direction arrows and the station dots and in order for them to be positioned in relation to their parent `<ul>` then I need that `relative` class on the `<ul>`

if there's a better/cleaner way to do this I'm all ears! (can also look into this more on my end)

js/components/ladderPage/ladder.tsx
cho-jos: That reasoning makes sense! I thought of another way to do this which is to just grab the styling string straight from the `Station` model -- although not sure if this is stylistically gucci from the code perspective. lmk any thoughts

js/components/ladderPage/ladder.tsx
mathcolo: Ah! You're right. I see in [this StackOverflow answer](https://stackoverflow.com/questions/12754159/in-css-position-is-absolute-inside-relative-absolute-to-the-document-or-the-pa) that:
> An element with position: absolute is absolute to it's nearest non-static parent container.

So, whoa. Nice!

Could you try removing `relative` on just the parent div?

mathcolo (commented):
Even before we do [🛠️ Implement swipe between ladders](https://app.asana.com/1/15492006741476/project/1209646113903402/task/1209815377576597), does this render left to right correctly on mobile?

js/data/stations.ts
```diff
@@ -0,0 +1,45 @@
...
+): Station => ({
+  id,
+  name,
+  marginBottom,
+  extraStyling,
+});
+
+export const StationSets = {
+  AlewifeAndrew: [
+    buildStation("place-alfcl", "Alewife", "mb-28", "pt-20"),
```
mathcolo, line 17: It's slightly strange for these otherwise ladder-specific CSS margin values to be here, a non-ladder-specific page. What you're really trying to represent (I think) in this data file is the ratios between stations, e.g. Park-DTX is 1 and DTX-South is 1.375? (I got those by dividing both by 8)

If this file contained those ratios, could you then try to escape from Tailwind (just for this specific case) and use `style={marginBottom: value}` to achieve the desired result?

cho-jos:
> Even before we do [🛠️ Implement swipe between ladders](https://app.asana.com/1/15492006741476/project/1209646113903402/task/1209815377576597), does this render left to right correctly on mobile?

I got the ladders to scroll horizontally via `overflow-x-auto` styling on their parent <div>, but now I noticed that the Orbit header is behaving weird on smaller screens (chrome device mode). The header just "ends" 
Screenshot 2025-04-28 at 4 34 23 PM

Tried playing around with the header styling, haven't figured it out yet

js/data/stations.ts
cho-jos: Agree with everything you said and reworked this. I actually like what it's doing now better, thanks for keeping Orbit (me) aligned with good coding practice -- lmk any thoughts on updated spacing calculation

mathcolo (approved):

PR #299, opened 2025-05-02
feat: Jcc attach pill to ladder

cho-jos:
Asana Task: [Attach pill to ladder](https://app.asana.com/1/15492006741476/project/1200273269966439/task/1209815377576595?focus=true)

demo trains include a pulsing Braintree train on main ladder, non-pulsing Ashmont train on Ashmont ladder
Screenshot 2025-05-02 at 5 07 15 PM

Checklist

- Tests:
  - `( )` Has tests
  - `(X)` Doesn't need tests
- Product/Design sign off:
  - `(X)` No review needed

### DIFF
--- a/js/components/ladderPage/ladder.tsx
+++ b/js/components/ladderPage/ladder.tsx
@@ -1,6 +1,7 @@
 import { StationSets } from "../../data/stations";
 import { useVehiclePositions } from "../../hooks/useVehiclePositions";
 import { Station } from "../../models/station";
+import { Train } from "./train";
 import { ReactElement } from "react";
 
 export const Ladder = (): ReactElement => {
@@ -19,12 +20,24 @@ export const Ladder = (): ReactElement => {
       </span>
       <div className="overflow-x-hidden">
         <div className="flex px-80 overflow-x-auto">
-          <StationList stations={StationSets.AlewifeAndrew} />
-          <StationList stations={StationSets.JFKAshmont} />
-          <StationList stations={StationSets.JFKBraintree} />
+          <TrainsAndStations />
         </div>
       </div>
     </>
+  );
+};
+
+// TODO: update to accept trains: Train[] parameter?
+const TrainsAndStations = (): ReactElement => {
+  return (
+    <div className="relative flex">
+      <StationList stations={StationSets.AlewifeAndrew} />
+      <StationList stations={StationSets.JFKAshmont} />
+      <StationList stations={StationSets.JFKBraintree} />
+      <div className="absolute top-[284px] left-[24px]">
+        <Train route="Red-Braintree" label="1888" direction={0} />
+      </div>
+    </div>
   );
 };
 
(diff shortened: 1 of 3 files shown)


js/components/ladderPage/train.tsx
```diff
@@ -5,26 +5,50 @@ export const Train = ({
...
+  direction,
   className: extraClassName,
 }: {
   route: "Red-Ashmont" | "Red-Braintree";
   label: string;
   highlight?: boolean;
+  direction: number;
   className?: string;
 }): ReactElement => {
+  const bgColor = route === "Red-Braintree" ? "bg-crimson" : "bg-tangerine";
```
cho-jos, line 17: is it safe to have an "else" case like this here? Since line 10/11 should ensure route is one of two possibilities?

js/components/ladderPage/ladder.tsx
```diff
@@ -19,15 +20,30 @@ export const Ladder = (): ReactElement => {
...
 };
 
+// TODO: update to accept trains: Train[] parameter?
+const TrainsAndStations = (): ReactElement => {
+  return (
+    <div className="relative flex">
+      <StationList stations={StationSets.AlewifeAndrew} />
+      <StationList stations={StationSets.JFKAshmont} />
+      <StationList stations={StationSets.JFKBraintree} />
+      <div className="absolute top-[284px] left-[24px]">
```
cho-jos, line 37: You can see just how arbitrarily I'm placing these trains on the ladder. I know for the purposes of this ticket it might be fine, but not sure if this actively makes downstream work more difficult or is lacking necessary structure

mathcolo: It should be fine, as we're not immediately concerned about overlapping 👍 Dynamic positioning will also probably, like the station names, need to be broken out of Tailwind

js/components/ladderPage/train.tsx
mathcolo: Safe to use the ternary, do you mean?

cho-jos: @mathcolo yes

mathcolo: Yep! that's fine

js/components/ladderPage/ladder.tsx
```diff
@@ -19,15 +20,35 @@ export const Ladder = (): ReactElement => {
...
+        <Train
+          route="Red-Braintree"
+          label="1888"
+          direction={0}
+          highlight={true}
+        />
+      </div>
+      <div className="absolute top-[316px] left-[704px]">
+        <Train route="Red-Ashmont" label="1889" direction={1} />
+      </div>
```
mathcolo, line 47: **question:** as I think about this more, would it make more sense for a single `TrainsAndStations` to render train and stations for a single ladder? Thinking about that in particular, as the `left` value of 704px is, right now, relative to the `relative` div on line 33. But that means the `left` values have to be different per ladder (as the div on line 33 covers the entire page), and it would be better for that not to be the case.

cho-jos: thinking about that conceptually, it might also make it easier in general to manage the future "heights" of the trains if we can narrow them down to the their specific ladder

mathcolo (approved):

PR #306, opened 2025-05-13
feat: interpolate location between stations

cho-jos:
Asana Task: [Interpolate location between stations](https://app.asana.com/1/15492006741476/project/1200273269966439/task/1209982475169316)

Checklist

- Tests:
  - `(X)` Has tests
  - `( )` Doesn't need tests
- Product/Design sign off:
  - `(X)` No review needed


js/components/ladderPage/ladder.tsx
```diff
@@ -20,37 +52,69 @@ export const Ladder = (): ReactElement => {
...
+        // TODO: this flat-out ignores that Ashmont-bound trains may be on the main ladder
+        const route =
+          stations.some((station) => station.id === "place-asmnl") ?
+            "Red-Ashmont"
+          : "Red-Braintree";
+
+        const direction =
+          vp.stopId && ["Alewife-02", "Braintree-02"].includes(vp.stopId) ? 0
+          : vp.stopId && ["Alewife-01", "Braintree-01"].includes(vp.stopId) ? 1
+          : vp.directionId;
```
mathcolo, line 101: **suggestion:** another case where we should probably prepare for BL and OL by parameterizing?

js/components/ladderPage/height.ts
```diff
@@ -0,0 +1,70 @@
+import { proportionBetweenLatLngs } from "../../models/latlng";
+import { Station } from "../../models/station";
+import { StopStatus, VehiclePosition } from "../../models/vehiclePosition";
+
+// TODO: index === -1 case?
+export const height = (pos: VehiclePosition, stationSet: Station[]) => {
+  const index = stationSet.findIndex((station) => pos.stationId === station.id);
+  let height = 68;
```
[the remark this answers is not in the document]
cho-jos: distance from the top of the ladder to the first station dot, accounting for the size of the pill

js/components/ladderPage/height.ts
```diff
@@ -0,0 +1,70 @@
...
+
+  // add up the bottom margins of stations up to index
+  // this will get us to the station that the vp is in relation to
+  for (let i = 0; i < index; i += 1) {
```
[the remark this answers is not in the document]
cho-jos: can you elaborate on this? what do you mean by "final" `stationId`?

js/components/ladderPage/height.ts
```diff
@@ -0,0 +1,70 @@
...
+    // proportionally backtrack progress if train is still in transit towards the station
+    if (pos.position !== null) {
+      travelLength = stationSet[index].spacingRatio * 32 + 24;
+      heightAdjustment =
+        proportionBetweenLatLngs(
+          stationSet[index + 1].location,
+          stationSet[index].location,
+          pos.position,
+        ) * travelLength;
+    }
```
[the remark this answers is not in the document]
cho-jos: @mathcolo lmk what you think about the new function(s), i still don't love it but my brain is wacked from trying to think of a different way to do this

js/components/ladderPage/height.ts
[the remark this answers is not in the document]
mathcolo: Ah! By "final" I mean `stationSet[index]`. `index` is computed on line 6, but the for loop on line 11 is another pass through the array. A `reduce` should be able to do this in one pass. (I'm not concerned about it from a performance perspective, but it might still be nice)

js/components/ladderPage/ladder.tsx
mathcolo: Maybe TrainsAndStations takes in a map of forced directions?
```
$ node
Welcome to Node.js v22.15.0.
Type ".help" for more information.
> const forced_directions = new Map([["Alewife-01", 1], ["Alewife-02", 0], ["Braintree-01", 1], ["Braintree-02", 0]])
undefined
> forced_directions.get("Alewife-01")
1
```

js/components/app.tsx
```diff
@@ -72,7 +73,15 @@ const router = createBrowserRouter([
       },
       {
         path: paths.ladder,
-        element: <Ladder />,
+        element: (
+          <Ladder
+            stationSets={[
```
mathcolo, line 78: **nit:** I forgot to mention this before (sorry!), but maybe this should be `StationList` instead of `StationSet` as it's not `Set` type, and ordering matters.

js/components/ladderPage/height.ts
```diff
@@ -0,0 +1,74 @@
+import { LatLng, proportionBetweenLatLngs } from "../../models/latlng";
+import { Station } from "../../models/station";
+import { StopStatus, VehiclePosition } from "../../models/vehiclePosition";
+
+export const height = (pos: VehiclePosition, stationSet: Station[]) => {
```
mathcolo, line 5: **question:** this function is nicely testable- could you throw a test suite at it?

js/components/ladderPage/height.ts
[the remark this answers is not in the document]
mathcolo: looks good! 🥳

js/components/ladderPage/height.ts
```diff
@@ -0,0 +1,74 @@
...
+  if (pos.stopStatus === StopStatus.StoppedAt) {
+    return height;
+  }
+
+  let start: LatLng = { latitude: 0.0, longitude: 0.0 };
+  const finish = stationSet[index].location;
+  let travelLength = 0.0;
+  if (pos.directionId === 0) {
+    // handle case where trains InTransitTo are "above" the first station
+    if (index === 0) {
```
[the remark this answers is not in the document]
cho-jos: I haven't noticed this situation arise. Thinking about it more I think the only "edge case" where we might specifically want to handle this is a northbound train traveling "above" Alewife (since northbound trains from JFK will jump to the main ladder). I can observe for this specifically and see what kind of behavior needs to be accounted for

js/components/ladderPage/height.ts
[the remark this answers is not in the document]
cho-jos: I was confused before but that makes sense. `reduce()` incoming

js/components/ladderPage/height.ts
```diff
@@ -0,0 +1,74 @@
...
+  }
+
+  if (pos.stopStatus === StopStatus.StoppedAt) {
+    return height;
+  }
+
+  let start: LatLng = { latitude: 0.0, longitude: 0.0 };
+  const finish = stationSet[index].location;
+  let travelLength = 0.0;
+  if (pos.directionId === 0) {
```
andrewdolce, line 28: **question:** I'm probably speaking from ignorance of the data, but are there human-readable names we can assign to `directionId` values of 0 and 1? From some of the documentation I've seen, it seems like the actual direction name varies by route? But since you seem to be able to make some assumptions based on 0 vs 1, I'm wondering if we could define a local enum or set of constants to compare against, at least for the purposes of upward vs downward direction along the ladder?

js/models/latlng.ts
```diff
@@ -9,3 +9,28 @@ export const LatLng = z.object({
...
+): number => {
+  const totalLatDelta: number = finish.latitude - start.latitude;
+  const totalLngDelta: number = finish.longitude - start.longitude;
+  const pointLatDelta: number = point.latitude - start.latitude;
+  const pointLngDelta: number = point.longitude - start.longitude;
+  const dotProductStartToPoint: number =
+    totalLatDelta * pointLatDelta + totalLngDelta * pointLngDelta;
+  const startToFinishMagnitudeSquared: number =
+    totalLatDelta * totalLatDelta + totalLngDelta * totalLngDelta;
+  return dotProductStartToPoint / startToFinishMagnitudeSquared;
```
andrewdolce: Do we want to guard against possible divide by zero here? In the event that `start` and `finish` are the same (arguably a misuse), this risks spitting out `NaN`. Maybe worth checking the denominator and returning `0` in that case.

andrewdolce: (Whoops - realizing this is in `main` already, so not part of this PR. Question still stands?)

js/components/ladderPage/height.ts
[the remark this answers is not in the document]
cho-jos: As discussed this morning, we actually landed at a single concise for loop instead of reduce()

js/components/ladderPage/height.ts
cho-jos: as far as I understand it, it's kind've yes and no... even though the different routes use `directionId`s of 0 and 1, it can refer to different things. On the red line we know they mean "southbound" vs. "northbound", but that isn't true for every line.  

I figured it's still safe to make assumptions based on 0 vs. 1 in this `height` function because when only concerning ourselves with the context of a train traveling along a visual ladder, there's only 2 ways to go (up or down the ladder). If/When Orbit eventually expands to also include Blue and Orange line ladders, that height calculation logic still works (hopefully?)

js/components/ladderPage/height.ts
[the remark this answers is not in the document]
cho-jos: @mathcolo I could be wrong (didn't observe the ladders long enough to notice a new edge case) but it _seems_ like consistently when a northbound train reaches Alewife, it instantly (or pretty close) updates to be `STOPPED_AT` one of the 2 track stopIds and visually in Orbit snaps to one of the open dots.

js/components/ladderPage/ladder.tsx
cho-jos: reworked to provide `TrainsAndStations` a map of `forcedDirections` if the route includes one (based off of reworking the `<Ladder />` to take in just a single `routeId` prop. 

Lmk any hesitations or dislikes

js/components/ladderPage/height.ts
```diff
@@ -0,0 +1,80 @@
...
+      break;
+    }
+    // spacingRatio * 32 = actual px value of bottom margin
+    // 24 = height of the station name
+    height += stationList[i].spacingRatio * 32 + 24;
+  }
+
+  // somehow the vp was not in the correct stationlist (unlikely)
+  if (index === -1) {
+    return -1;
```
mathcolo, line 22: **suggestion:** maybe console.error this so we have a chance at noticing it?

js/components/ladderPage/height.ts
```diff
@@ -0,0 +1,80 @@
...
+  let start: LatLng = { latitude: 0.0, longitude: 0.0 };
+  const finish = stationList[index].location;
+  let travelLength = 0.0;
+  if (pos.directionId === 0) {
+    // handle case where trains InTransitTo are "above" the first station
+    if (index === 0) {
+      return 20;
+    }
+    start = stationList[index - 1].location;
+    travelLength = stationList[index - 1].spacingRatio * 32 + 24;
```
mathcolo, line 41: **nit:** should`stationList[index - 1]` be broken out into a local variable?

js/components/ladderPage/ladder.tsx
```diff
@@ -1,13 +1,59 @@
...
   const tripUpdates = useTripUpdates();
+  const vehiclePositions = useVehiclePositions();
+  const stationLists: Station[][] =
+    routeId === "Red" ?
+      [
+        RedStationLists.AlewifeAndrew,
+        RedStationLists.JFKAshmont,
+        RedStationLists.JFKBraintree,
+      ]
+    : [];
```
mathcolo, line 20: **suggestion:** this is a little strange—could it instead be a constant in another file, or something? Thinking along the lines of:
```
const stationLists = {
  Red: [
      AlewifeAndrew,
      JFKAshmont,
      JFKBraintree
  ],
   [...]
```

js/components/ladderPage/ladder.tsx
```diff
@@ -1,13 +1,59 @@
...
+
+  const forcedDirections =
+    routeId === "Red" ?
+      new Map([
+        ["Alewife-01", 1],
+        ["Alewife-02", 0],
+        ["Braintree-01", 1],
+        ["Braintree-02", 0],
```
mathcolo, line 28: **suggestion**: akin to above, perhaps all Red Line configuration you need for this to function can go in one place?

js/components/ladderPage/ladder.tsx
```diff
@@ -25,37 +71,71 @@ export const Ladder = (): ReactElement => {
...
 const TrainsAndStations = ({
   stations,
-  // vps
+  vps,
+  forcedDirections,
 }: {
   stations: Station[];
-  // vps: VehiclePosition[]
+  vps: VehiclePosition[];
+  forcedDirections: Map<string, number> | null;
```
mathcolo, line 98: **suggestion:** since there are likely to be `forcedDirections` for every line, maybe this shouldn't be null? Then the optional chaining operators on 125 could be removed.

js/test/components/ladderPage/height.test.tsx
```diff
@@ -0,0 +1,58 @@
...
+    });
+
+    test("for a valid VehiclePosition past Quincy Adams", () => {
+      const pos = vehiclePositionFactory.build({
+        stationId: "place-brntn",
+        stopId: "Braintree-01",
+        stopStatus: StopStatus.StoppedAt,
+      });
+      expect(height(pos, RedStationLists.JFKBraintree)).toBe(900);
+    });
```
mathcolo, line 29: **suggestion:** I know you have a test for southbound below (where above Alewife), but perhaps we should test a regular southbound case as well?

js/components/ladderPage/height.ts
cho-jos: ah I thought to [capture](https://github.com/mbta/orbit/blob/pim-dynamic-height/js/components/ladderPage/ladder.tsx#L112) this in `ladder.tsx` actually

js/data/stations.ts
```diff
@@ -1,43 +1,180 @@
...
-  name: string,
-  spacingRatio: number,
-): Station => ({
-  id,
-  name,
-  spacingRatio,
-});
-
-export const StationSets = {
+export const RedStationLists = {
```
mathcolo, line 1: **suggestion:** if this was
```
export const StationLists = {
    Red: {
        AlewifeAndrew: [
            [...]
```

maybe `routes.ts` doesn't need to exist?

js/components/ladderPage/ladder.tsx
```diff
@@ -1,13 +1,44 @@
...
 import { Train } from "./train";
 import { ReactElement } from "react";
 
-export const Ladder = (): ReactElement => {
-  const vehiclePositions = useVehiclePositions();
+export const Ladder = ({ routeId }: { routeId: string }): ReactElement => {
   const tripUpdates = useTripUpdates();
+  const vehiclePositions = useVehiclePositions();
+  const stationLists: Station[][] = routeId === "Red" ? routeLists.Red : [];
+  const forcedDirections = routeId === "Red" ? trackSides.Red : null;
```
mathcolo, line 15: **suggestion:** can you look up the data in `routeLists` or `trackSides` using the `routeId` that comes in as the argument? or change `routeId`'s TypeScript type to require it to be "Red" for now?

js/components/ladderPage/ladder.tsx
```diff
@@ -1,13 +1,44 @@
...
 import { useVehiclePositions } from "../../hooks/useVehiclePositions";
 import { Station } from "../../models/station";
+import { StopStatus, VehiclePosition } from "../../models/vehiclePosition";
+import { height } from "./height";
 import { Train } from "./train";
 import { ReactElement } from "react";
 
-export const Ladder = (): ReactElement => {
-  const vehiclePositions = useVehiclePositions();
+export const Ladder = ({ routeId }: { routeId: string }): ReactElement => {
```
mathcolo, line 11: **suggestion:** prefer using the `routeId` type in `./common`

js/components/ladderPage/ladder.tsx
```diff
@@ -25,37 +56,71 @@ export const Ladder = (): ReactElement => {
...
     <div className="relative flex">
       <StationList stations={stations} />
+      {vps.map((vp) => {
+        if (vp.position === null && vp.stopStatus !== StopStatus.StoppedAt) {
+          return null;
+        }
+
+        const trainHeight = height(vp, stations);
+        // add 80 for top margin above the station list borders
+        const px = trainHeight == -1 ? -1 : trainHeight + 80;
```
mathcolo, line 95: **suggestion:** prefer triple equals (===) instead of double equals (==) ([explanation](https://stackoverflow.com/a/60669874))

js/components/ladderPage/ladder.tsx
```diff
@@ -25,37 +56,71 @@ export const Ladder = (): ReactElement => {
...
+
+        // TODO: this flat-out ignores that Ashmont-bound trains may be on the main ladder
+        const route =
+          stations.some((station) => station.id === "place-asmnl") ?
+            "Red-Ashmont"
+          : "Red-Braintree";
+
+        const direction: number =
+          vp.stopId && forcedDirections?.has(vp.stopId) ?
+            (forcedDirections.get(vp.stopId) ?? vp.directionId)
```
mathcolo, line 111: **suggestion:** `.get` returns `undefined` if it isn't present, so you shouldn't need the `has`?

cho-jos: if I remove the check with `.has()` then there are downstream complaints in `<Train />` that direction may be `undefined` even though it is specifically typed to `direction: number;`. Using `has()` and the nullish coalesce was the quickest way I found to get the errors to quiet down, but if there's a smarter way to handle this I'm all ears

js/data/stations.ts
cho-jos: ^this is actually more intuitive to me, but I'm a typescript noob and idk how/what type to give `StationList.Red` to `<Ladder />` to use as an array of station arrays. Any pointers?

js/components/ladderPage/ladder.tsx
cho-jos: > can you look up the data in routeLists or trackSides using the routeId that comes in as the argument?

how do I access an object's (struct's?) key using a string?

mathcolo: Object is right :-) You can do `someObject[someString]`. If and only if the string is hardcoded, you can use dot-access syntax such as `someObject.Red`.

js/data/stations.ts
mathcolo: Probably a [Record](https://www.typescriptlang.org/docs/handbook/utility-types.html#recordkeys-type) of Records, actually!

Maybe `Record<RouteId, Record<string, StationConfiguration>`?

js/components/ladderPage/height.ts
```diff
@@ -0,0 +1,82 @@
...
+
+  // somehow the vp was not in the correct stationlist (unlikely)
+  if (index === -1) {
+    return -1;
+  }
+
+  const currentStation = stationList[index];
+  if (currentStation.id === "place-brntn") {
+    height += 24; // extra padding because "Quincy Adams" wraps
+  }
```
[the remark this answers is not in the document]
cho-jos: hmm... your point is valid, however as far as I understand since Quincy Adams is the only station this applies to and it's very much a specific isolated edge-case I'd like to explicitly handle it in the height function as opposed to adding another optional/one-off field to `Station`. Thoughts?

js/components/ladderPage/ladder.tsx
```diff
@@ -1,13 +1,38 @@
-import { StationSets } from "../../data/stations";
+import { Ladder, Stations } from "../../data/stations";
```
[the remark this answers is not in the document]
mathcolo: yep 👍

js/components/ladderPage/height.ts
```diff
@@ -0,0 +1,82 @@
...
+  let start: LatLng = { latitude: 0.0, longitude: 0.0 };
+  const finish = currentStation.location;
+  let travelLength = 0.0;
+  if (pos.directionId === 0) {
+    // handle case where trains InTransitTo are "above" the first station
+    if (index === 0) {
+      return 20;
+    }
+    const prevStation = stationList[index - 1];
+    start = prevStation.location;
```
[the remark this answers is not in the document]
cho-jos: @mathcolo this ultimately lead to a slight rewrite of `height()`. I hope it's clearer, easier on the eyes, and perhaps more Typescripty

js/components/ladderPage/ladder.tsx
```diff
@@ -25,42 +50,82 @@ export const Ladder = (): ReactElement => {
...
-              label="1888"
-              direction={0}
-              highlight={true}
-            />
-          </div>
-          <div className="absolute top-[316px] left-[1024px]">
-            <Train route="Red-Ashmont" label="1889" direction={1} />
-          </div>
+          {vpsByBranch &&
+            Array.from(vpsByBranch.entries()).map(
```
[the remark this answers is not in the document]
mathcolo: ah, I see 👍 got it!

js/components/ladderPage/ladder.tsx
```diff
@@ -25,42 +50,86 @@ export const Ladder = (): ReactElement => {
...
+              ([StationLists, vps], index) => (
+                <TrainsAndStations
+                  key={index}
+                  stations={StationLists}
```
mathcolo, line 58: **nit:** is this really a `config`, not `stations` now?

js/components/ladderPage/ladder.tsx
```diff
@@ -25,42 +50,86 @@ export const Ladder = (): ReactElement => {
...
 
-// TODO: accept VehiclePositions and use them to determine the <Train />'s to render
 const TrainsAndStations = ({
   stations,
-  // vps
+  vps,
 }: {
-  stations: Station[];
-  // vps: VehiclePosition[]
+  stations: LadderConfig;
```
mathcolo, line 73: nit: same as above

js/models/latlng.ts
```diff
@@ -32,5 +32,7 @@ export const proportionBetweenLatLngs = (
     totalLatDelta * pointLatDelta + totalLngDelta * pointLngDelta;
   const startToFinishMagnitudeSquared: number =
     totalLatDelta * totalLatDelta + totalLngDelta * totalLngDelta;
-  return dotProductStartToPoint / startToFinishMagnitudeSquared;
+  return startToFinishMagnitudeSquared === 0.0 ? 0.0 : (
```
mathcolo, line 35: **question:** ?

cho-jos: This was (perhaps a bad) attempt to guard against 0 division as suggested by @andrewdolce

js/components/ladderPage/ladder.tsx
cho-jos: how about making it even clearer/Orbit specific with `ladderConfig` >

mathcolo (commented):
super close!

js/components/ladderPage/ladder.tsx
```diff
@@ -25,42 +50,93 @@ export const Ladder = (): ReactElement => {
...
+        const route =
+          ladderConfig.some((station) => station.id === "place-asmnl") ?
+            "Red-Ashmont"
+          : "Red-Braintree";
+
+        const stationWithForcedDirection = ladderConfig.find(
+          (station) =>
+            station.forcedDirections !== undefined &&
+            station.stop_ids.some((stop_id) => stop_id === vp.stopId),
+        );
```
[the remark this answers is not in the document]
mathcolo: Could you do:

```
const station = ladderConfig.find((station) =>
   station.stop_ids.some((stop_id) => stop_id === vp.stopId),
);
```

and then later, this?

```
const direction = station.forcedDirections?.get(vp.stopId) ?? vp.directionId;
```

js/components/ladderPage/height.ts
```diff
@@ -0,0 +1,100 @@
...
+  if (pos.directionId === 0 && index === 0) {
+    return 20;
+  } else if (pos.directionId === 1 && index === stationList.length - 1) {
+    // handle case where trains InTransitTo are "below" the first station
+    return height + 40;
+  }
+
+  const start =
+    pos.directionId === 0 ?
+      stationList[index - 1].location
```
mathcolo, line 45: **question:** just verifying that `index - 1` is never going to go negative?

cho-jos: @mathcolo logically speaking, it should be impossible

if index = -1, we handle that above this line and return before hitting it.
if the vp is `StoppedAt` (regardless of the index, but includes when index = 0) we handle that above this line and return before hitting it.
If the vp is still `InTransitTo` moving southbound and the index = 0, we handle handle that specifically above this line and return before hitting it.

if the vp is moving northbound then we never compute `index - 1`, only `index + 1` which is guarded against going out of bounds because of northbound versions of the logic mentioned above, especially capturing northbound trains that are `InTransitTo` with an index that is `stationList.length - 1` and handling that specially.

PR #323, opened 2025-05-30
feat: render train pill colors based on Ashmont/Braintree

andrewdolce:
Asana Task: [Ladder pill colors based on Ashmont/Braintree](https://app.asana.com/1/15492006741476/project/1200273269966439/task/1210239106758138?focus=true)

Displays correct colors for train pills on the ladder, with Ashmont trains shown in orange, and Braintree trains shown in red. This determination is made based on the `route_pattern_id` of the vehicle's current `TripUpdate`.

Note that the AC on the ticket says to apply these colors "regardless of ladder", but there was past discussion about whether there could be cases where "Ashmont" trains (based on route pattern) are briefly on the Braintree ladder, or vice versa. So for now I've left in the ladder-based logic and am only using the pattern ID for the Alewife branch. (I'll call this out in the code).

Screenshot 2025-05-30 at 12 00 46 PM

Couple other points of discussion:

- Amended the backend data model for `VehiclePosition` with the minimum change needed to forward the trip descriptor (ie, thing that contains the trip ID). Then matching the vehicle position with trip updates is done on the frontend. Preston and I had talked about alternatively doing that work on the backend (maybe using ecto to store an actual relationship between the two?). I'm still open to that but wanted to get this smaller change up first.

- I created a frontend-only `Vehicle` model, which represents a vehicle position that has (potentially) been matched with a trip update. I refactored ladder and train views to use this new model. Definitely open to feedback on the details here. (I am also considering defining a `useVehicles` hook that does the matching outside of the ladder component.)

- I need to touch base with transit data team to double check that we're not abusing the data in ways that will break, and also to verify little details, like double checking nullability / presence / absence of the trip ID. For example, the docs on the vehicle position's [TripDescriptor](https://github.com/google/transit/blob/master/gtfs-realtime/spec/en/reference.md#message-tripdescriptor) explain that trip ID is not guaranteed. Depending on whether that actually happens for orbit trips, we may need to follow up with some other way of matching vehicles to trip updates.

Checklist

- Tests:
  - `(X)` Has tests
  - `( )` Doesn't need tests
- Product/Design sign off:
  - `( )` No review needed

### DIFF
--- a/js/components/ladderPage/ladder.tsx
+++ b/js/components/ladderPage/ladder.tsx
@@ -114,10 +114,13 @@ const TrainsAndStations = ({
         const isAshmontLadder = ladderConfig.some(
           (station) => station.id === "place-asmnl",
         );
+        const isBraintreeLadder = ladderConfig.some(
+          (station) => station.id === "place-brntn",
+        );
         const routePattern =
-          isAshmontLadder ? "Red-Ashmont" : (
-            (trainRoutePatternFromVehicle(vehicle) ?? "Red-Braintree")
-          );
+          isAshmontLadder ? "Red-Ashmont"
+          : isBraintreeLadder ? "Red-Braintree"
+          : (trainRoutePatternFromVehicle(vehicle) ?? "Red-Braintree");
 
         const station = ladderConfig.find((station) =>
           station.stop_ids.some((stop_id) => stop_id === vp.stopId),
--- a/js/components/ladderPage/ladder.tsx
+++ b/js/components/ladderPage/ladder.tsx
@@ -3,6 +3,7 @@ import { useTripUpdates } from "../../hooks/useTripUpdates";
 import { useVehiclePositions } from "../../hooks/useVehiclePositions";
 import { RouteId } from "../../models/common";
 import {
+  trainRoutePatternFromVehicle,
   Vehicle,
   vehiclesFromPositionsAndTripUpdates,
 } from "../../models/vehicle";
@@ -108,11 +109,15 @@ const TrainsAndStations = ({
         // add 80 for top margin above the station list borders
         const px = trainHeight + 80;
 
-        // TODO: this flat-out ignores that Ashmont-bound trains may be on the main ladder
-        const route =
-          ladderConfig.some((station) => station.id === "place-asmnl") ?
-            "Red-Ashmont"
-          : "Red-Braintree";
+        // TODO: Discuss if we actually want to keep the ladder-based logic here, or just always
+        // rely on the pattern name.
+        const isAshmontLadder = ladderConfig.some(
+          (station) => station.id === "place-asmnl",
+        );
+        const routePattern =
+          isAshmontLadder ? "Red-Ashmont" : (
+            (trainRoutePatternFromVehicle(vehicle) ?? "Red-Braintree")
+          );
 
         const station = ladderConfig.find((station) =>
           station.stop_ids.some((stop_id) => stop_id === vp.stopId),
@@ -129,7 +134,11 @@ const TrainsAndStations = ({
             style={{ position: "absolute", top: `${px}px` }}
             className={direction === 0 ? "left-[24px]" : "right-[24px]"}
           >
-            <Train route={route} label={vp.label} direction={direction} />
+            <Train
+              routePattern={routePattern}
+              label={vp.label}
+              direction={direction}
+            />
           </div>
         );
       })}
--- a/js/components/ladderPage/ladder.tsx
+++ b/js/components/ladderPage/ladder.tsx
@@ -2,7 +2,11 @@ import { LadderConfig, Stations } from "../../data/stations";
 import { useTripUpdates } from "../../hooks/useTripUpdates";
 import { useVehiclePositions } from "../../hooks/useVehiclePositions";
 import { RouteId } from "../../models/common";
-import { StopStatus, VehiclePosition } from "../../models/vehiclePosition";
+import {
+  Vehicle,
+  vehiclesFromPositionsAndTripUpdates,
+} from "../../models/vehicle";
+import { StopStatus } from "../../models/vehiclePosition";
 import { height } from "./height";
 import { Train } from "./train";
 import { ReactElement } from "react";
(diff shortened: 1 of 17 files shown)


lib/realtime/data/vehicle_position.ex
```diff
@@ -17,7 +17,7 @@ defmodule Realtime.Data.VehiclePosition do
           timestamp: DateTime.t(),
           # GTFS-RT id, e.g. "R-54826B16".
           vehicle_id: String.t() | nil,
-          trip: nil
+          trip: Realtime.Data.TripDescriptor.t()
```
andrewdolce, line 20: **question**: Something I definitely want feedback on: I see this pre-existing (but not yet used) `trip` field. Right now I have it repurposed to hold the `TripDescriptor` struct, to match with the GTFS realtime docs.

But maybe the intention was for this to someday represent a link in the ORM to an actual trip update entity? If so, happy to change it.

Basically I've thought of 3 options:

1. Leave it as, with `trip` storing a `TripDescriptor`.
2. Instead, give the trip descriptor its own separate `tripDescriptor` field, and reserve `trip` for later.
3. If we decide that we actually only care about the trip ID, and not any of the other things that we might want to add to `TripDescriptor`, I can instead just add a `trip_id` field, like we have for `vehicle_id`.

Let me know if there are strong opinions.

lib/realtime/data/trip_descriptor.ex
```diff
@@ -0,0 +1,54 @@
...
+          trip_id: String
+        }
+  @derive {Jason.Encoder,
+           only: [
+             :trip_id
+           ]}
+
+  defstruct [
+    :trip_id
+  ]
```
andrewdolce, line 22: Note that TripDescriptor can have other fields. For now I'm only storing the trip ID, but I figured I'd set this up so there is an obvious place to add should we care about the others.

js/components/ladderPage/ladder.tsx
```diff
@@ -98,11 +109,18 @@ const TrainsAndStations = ({
...
+        const isAshmontLadder = ladderConfig.some(
+          (station) => station.id === "place-asmnl",
+        );
+        const isBraintreeLadder = ladderConfig.some(
+          (station) => station.id === "place-brntn",
+        );
+        const routePattern =
+          isAshmontLadder ? "Red-Ashmont"
+          : isBraintreeLadder ? "Red-Braintree"
+          : (trainRoutePatternFromVehicle(vehicle) ?? "Red-Braintree");
```
andrewdolce, line 123: **question**: I know we talked about this, but do we still want to check based on ladder, like I'm doing here? Also note that in the event that we cannot determine a route pattern from the vehicle, I'm defaulting to "Braintree".

lib/realtime/data/vehicle_position.ex
```diff
@@ -17,7 +17,7 @@ defmodule Realtime.Data.VehiclePosition do
           timestamp: DateTime.t(),
           # GTFS-RT id, e.g. "R-54826B16".
           vehicle_id: String.t() | nil,
-          trip: nil
+          trip_id: String.t() | nil
```
andrewdolce, line 20: After deliberating (see my [outdated comments](https://github.com/mbta/orbit/pull/323#discussion_r2121604918)) I'm opting to just store the trip ID, mapped over from the vehicle position's `TripDescriptor`, since, after checking with Eddie, it turns out that (for the red line trains serving a current trip) the trip ID is always expected to exist, and we don't need other matching criteria, at least within a given service date.

If we want to keep the unused `trip` field for future use, let me know.

js/components/ladderPage/ladder.tsx
mathcolo: +1 to @cho-jos. It shouldn't matter too much, but it's probably best to trust RTR first and fall back only if route pattern isn't present

js/models/trainRoutePattern.ts
```diff
@@ -0,0 +1,17 @@
+// Represents a hard-coded list of understood route patterns
+export type TrainRoutePattern = "Red-Ashmont" | "Red-Braintree";
```
mathcolo, line 2: **question (non-blocking):** This is slightly odd in that "Red-Ashmont" and "Red-Braintree" don't actually exist from a GTFS perspective—they're not actual route patterns. Wondering if this file would be better set up as a mapping from route pattern to color, and called `colors.ts` or similar?

js/models/vehicle.ts
```diff
@@ -0,0 +1,44 @@
...
+    }
+    tripUpdateById.set(tripUpdate.tripId, tripUpdate);
+  }
+
+  return vehiclePositions.map((vehiclePosition) => {
+    const tripId = vehiclePosition.tripId;
+    const tripUpdate = tripId && tripUpdateById.get(tripId);
+    return {
+      vehiclePosition,
+      ...(tripUpdate ? { tripUpdate } : {}),
```
mathcolo, line 32: **suggestion:** `Map.get` returns `undefined` if the key doesn't exist, so can this become
```
return {
  vehiclePosition,
  tripUpdate: tripUpdateById.get(tripId),
}
```
?

lib/realtime/data/trip_descriptor.ex
mathcolo: This is good!

My only wondering is about OCS trip data. If we retained the "just speak GTFS" feel of the current client-server communication, perhaps we could issue OCS-origin TripUpdates alongside RTR-origin TripUpdates? That might be in line with the fresh but growing longer term objective of having RTR make predictions but get OCS data from Orbit or something else upstream.

js/components/ladderPage/train.tsx
```diff
@@ -1,20 +1,22 @@
+import { TrainRoutePattern } from "../../models/trainRoutePattern";
 import { className } from "../../util/dom";
 import { ReactElement } from "react";
 
 export const Train = ({
-  route,
+  routePattern,
```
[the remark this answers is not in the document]
andrewdolce: Yeah I agree with that. The Train doesn't need to be smart enough to know about routes, and it'll be easier to extend later when we add orange/blue, or for other edge cases where design might decide we want special coloring.

js/models/vehicle.ts
andrewdolce: So in the past I've had a preference for not including keys with an explicitly `undefined` value, since in JS these are _technically_ different things (and IIRC, conform to different type definitions in typescript):

```
{ foo: 5, bar: undefined }
{ foo: 5 }
```
Not a sticking point for me though, so happy to change it.

mathcolo: Ah, gotcha. Is the explicitness of that key actually a feature, not a bug? 😅 

Indifferent knowing that, though.

lib/realtime/data/trip_descriptor.ex
mathcolo: @andrewdolce Oh, no, I actually wasn't prescribing a detail at that level—more just thinking broadly about where OCS data will fit in. All good by me 👍 we can change it later if we need to.

cho-jos:
I like it. Personally ready to approve but see there might still be some unresolved threads w/ Preston. Don't want to supersede him and approve if conversations are still happening.

js/models/vehicle.ts
```diff
@@ -0,0 +1,31 @@
...
+export const vehiclesFromPositionsAndTripUpdates = (
+  vehiclePositions: VehiclePosition[],
+  tripUpdates: TripUpdate[],
+): Vehicle[] => {
+  const tripUpdateById = new Map<string, TripUpdate>();
+  for (const tripUpdate of tripUpdates) {
+    const existing = tripUpdateById.get(tripUpdate.tripId);
+    if (existing) {
+      console.error("Multiple updates for trip. Overwriting");
+      console.error(existing);
```
mathcolo, line 18: **nit:** could these two console.errors be merged into one, so they never become separated from each other amid other output/errors in the worst case?

mathcolo: 👍 fine to remove!

js/components/ladderPage/ladder.tsx
```diff
@@ -68,15 +82,34 @@ export const Ladders = ({ routeId }: { routeId: RouteId }): ReactElement => {
...
 const TrainsAndStations = ({
   ladderConfig,
-  vps,
+  vehicles,
 }: {
   ladderConfig: LadderConfig;
-  vps: VehiclePosition[];
+  vehicles: Vehicle[];
 }): ReactElement => {
+  const themeForVehicle = (vehicle: Vehicle): TrainTheme => {
```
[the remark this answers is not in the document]
andrewdolce: If I'm following, I'll move it out of this component and pass in the ladder config as parameter? Thinking I'll keep it inside the `ladder.tsx` module for now (until it maybe becomes useful in other places?). Otherwise I'd probably want to create yet another module, since it doesn't feel like it can go solely in `vehicle.ts`, `trainThemes.ts`, or `stations.ts` without making the dependencies more convoluted.

PR #331, opened 2025-06-13
fix(trip_updates): handle null fields in trip_update parsing

andrewdolce:
Asana Task: [Crash due to non-nullable TripUpdates fields](https://app.asana.com/1/15492006741476/project/1200882337457260/task/1210526410605224?focus=true)

This is a quick fix for the null direction crash. I haven't fully sifted through the trip updates JSON and the GTFS spec, but it looks like there are cases where `direction_id` is not provided, and the orbit backend ends up mapping that into a value of `direction: nil`. The orbit frontend expects this to be non-null, and so zod crashes during parsing.

Whether these values should be allowed in the long term is a bigger question, but right now we're not using them, so I think it's worth merging this in the short-term.

Checklist

- Tests:
  - `( )` Has tests
  - `( )` Doesn't need tests
  - `(X)` Tests deferred (with justification)
- Product/Design sign off:
  - `(X)` No review needed

### DIFF
--- a/js/models/tripUpdate.ts
+++ b/js/models/tripUpdate.ts
@@ -16,8 +16,8 @@ export type TripUpdate = {
   direction: number | null;
   routePatternId: string | null;
   tripId: string;
-  vehicleId: string;
-  timestamp: DateTime;
+  vehicleId: string | null;
+  timestamp: DateTime | null;
   stopTimeUpdates: StopTimeUpdate[];
 };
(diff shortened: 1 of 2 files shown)


mathcolo:
Discussed in Slack—should be good if we add vehicle ID and timestamp. (Other thought: should we filter out `timestamp==null`? :/)

PR #335, opened 2025-06-18
feat(ocs): add ocs_trips and ocs_trains tables

mathcolo:
Asana Task: [🛠 Set up data structures for OCS trainsheets](https://app.asana.com/1/15492006741476/project/1206105669438487/task/1210389787136611)

Create two new tables to populate with OCS data as we receive it:
1. ocs_trips
2. ocs_trains

NB: I'm not using foreign key constraints here, as I don't have enough certainty about the order of insertion. So I'm expecting using this data to be best-effort.

Checklist

- Tests:
  - `(x)` Has tests
  - `( )` Doesn't need tests
- Product/Design sign off:
  - `( )` No review needed

### DIFF
--- a/lib/orbit/ocs/train.ex
+++ b/lib/orbit/ocs/train.ex
@@ -8,8 +8,9 @@ defmodule Orbit.Ocs.Train do
           service_date: Date.t(),
           uid: String.t(),
           rail_line: RailLine.t(),
-          cars: [String.t()] | nil,
           tags: [String.t()] | nil,
+          cars: [String.t()] | nil,
+          car_tags: [String.t()] | nil,
           deleted: boolean() | nil
         }
   schema "ocs_trains" do
@@ -17,8 +18,9 @@ defmodule Orbit.Ocs.Train do
     field(:uid, :string)
     field(:rail_line, RailLine)
 
-    field(:cars, {:array, :string})
     field(:tags, {:array, :string})
+    field(:cars, {:array, :string})
+    field(:car_tags, {:array, :string})
     field(:deleted, :boolean)
 
     timestamps()
@@ -32,8 +34,9 @@ defmodule Orbit.Ocs.Train do
         :service_date,
         :uid,
         :rail_line,
-        :cars,
         :tags,
+        :cars,
+        :car_tags,
         :deleted
       ]
     )
--- a/lib/orbit/ocs/train.ex
+++ b/lib/orbit/ocs/train.ex
@@ -1,0 +1,47 @@
+defmodule Orbit.Ocs.Train do
+  use Ecto.Schema
+  alias Orbit.RailLine
+
+  import Ecto.Changeset
+
+  @type t :: %__MODULE__{
+          service_date: Date.t(),
+          uid: String.t(),
+          rail_line: RailLine.t(),
+          cars: [String.t()] | nil,
+          tags: [String.t()] | nil,
+          deleted: boolean() | nil
+        }
+  schema "ocs_trains" do
+    field(:service_date, :date)
+    field(:uid, :string)
+    field(:rail_line, RailLine)
+
+    field(:cars, {:array, :string})
+    field(:tags, {:array, :string})
+    field(:deleted, :boolean)
+
+    timestamps()
+  end
+
+  def changeset(struct, attrs \\ %{}) do
+    struct
+    |> cast(
+      attrs,
+      [
+        :service_date,
+        :uid,
+        :rail_line,
+        :cars,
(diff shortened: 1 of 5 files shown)


lib/orbit/ocs/train.ex
```diff
@@ -0,0 +1,47 @@
+defmodule Orbit.Ocs.Train do
+  use Ecto.Schema
+  alias Orbit.RailLine
+
+  import Ecto.Changeset
+
+  @type t :: %__MODULE__{
+          service_date: Date.t(),
+          uid: String.t(),
```
andrewdolce, line 9: **question**: Will this UID match the internal "Train UID" used in OCS (not guaranteed to be globally unique outside of a given transit line)? Or something that we will generate unique to Orbit's DB?

Also do we think it's worth explicitly storing the `TMOV` "Train ID" field? It's potentially redundant (ie, just the lead car number + up to two tags prefixed), but in cases where there are more than 2 tags, reconstructing it could be ambiguous.

priv/repo/migrations/20250617182608_ocs.exs
```diff
@@ -0,0 +1,39 @@
...
+      add(:scheduled_arrival, :utc_datetime)
+      add(:origin_station, :string)
+      add(:destination_station, :string)
+      add(:deleted, :boolean)
+
+      timestamps()
+    end
+
+    create unique_index(:ocs_trips, [:service_date, :uid, :rail_line])
+    create index(:ocs_trips, [:service_date, :train_uid])
```
andrewdolce, line 24: **question**: Would it also make sense to also include the service date in this index, since you can't really uniquely specify a train without that?

mathcolo: It's there, am I missing something?

andrewdolce: Err, sorry. I typed "service date" but I meant "rail line".

lib/orbit/ocs/trip.ex
```diff
@@ -0,0 +1,71 @@
...
+          service_date: Date.t(),
+          uid: String.t(),
+          train_uid: String.t() | nil,
+          prev_uid: String.t() | nil,
+          next_uid: String.t() | nil,
+          route: String.t() | nil,
+          rail_line: RailLine.t(),
+          trip_type: String.t() | nil,
+          scheduled_departure: DateTime.t() | nil,
+          scheduled_arrival: DateTime.t() | nil,
```
andrewdolce, line 17: Need a field for `offset`?

mathcolo: thanks! 746ad71755ca7944f5b9f50d343f2472f316b04d

lib/orbit/ocs/train.ex
andrewdolce: Well, thinking on it more, I don't think we care about it.

But to answer the question, from the OCS TMOV docs:

```
-   *Eleventh*: Train ID -- 6 characters representing, 2 tags and the
    lead car number.
```

We can already get the lead car number and the full set of tags via other means, so we'll already have that information. The only reason we would care to store the explicit train ID is if said ID were needed/useful for some other purpose... but I don't know that there is one, so probably not worth worrying about it for now.

lib/orbit/ocs/train.ex
```diff
@@ -0,0 +1,47 @@
...
+          rail_line: RailLine.t(),
+          cars: [String.t()] | nil,
+          tags: [String.t()] | nil,
```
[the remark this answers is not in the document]
andrewdolce: Based on slack conversation, sounds like this is all we need, unless we decide we care about the tags on a per-car basis.

lib/orbit/ocs/train.ex
```diff
@@ -0,0 +1,47 @@
...
+  use Ecto.Schema
+  alias Orbit.RailLine
+
+  import Ecto.Changeset
+
+  @type t :: %__MODULE__{
+          service_date: Date.t(),
+          uid: String.t(),
+          rail_line: RailLine.t(),
+          cars: [String.t()] | nil,
```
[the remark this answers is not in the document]
andrewdolce: Hmm, I guess not.

lib/orbit/ocs/train.ex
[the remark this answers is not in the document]
mathcolo: This doesn't preclude per-car tags, does it?

I'm imagining using this field to store an array of tags with the same length as `cars`.

PR #336, opened 2025-06-18
feat(ocs): set up ocs.raw_message kinesis consumer 

andrewdolce:
Asana Task: [Have Orbit read in the new TSCH messages from Trike](https://app.asana.com/1/15492006741476/project/1200273269966439/task/1209929021017070?focus=true)

This is only part 1 of the work for getting kinesis messages.

- Checks in kinesis consumer boilerplate code provided by broadway_kinesis. Note that _for now_ I have copied this directly into our codebase, rather than include it as a mix dependency, as it is not yet published in hex (and including it it via `:git` was causing issues in my local dev setup.) If we decide to actually use the library, I will remove this in favor of an actual dependency via mix.
- Checks in OCS-specific stream code, adapted from RTR. This does not yet include the OCS message models or parser, which I plan to add in a future PR.
- Makes some configuration changes - Notably this introduces a requirement that we add some secrets in `dev.secret.exs`. I indicated this in the example file, but still need to round out the README.

Still TODO before this is merged:

- Add config notes to the README
- Add persistent state to the kinesis producer, like RTR has. (Might defer to future PR).
- Tests

Checklist

- Tests:
  - `( )` Has tests
  - `( )` Doesn't need tests
- Product/Design sign off:
  - `( )` No review needed


docker-compose.yml
```diff
@@ -28,5 +28,7 @@ services:
       - S3_RTR_PUBLIC=s3-anonymous://mbta-gtfs-s3/
       - GLIDES_PERSONNEL_PATHNAME=personnel/demo.csv
       - GLIDES_RFID_PATHNAME=rfid/demo.csv
+      - OCS_KINSESIS_STREAM_NAME=ctd-ocs-raw-messages
```
mathcolo, line 31: nit: typo

lib/ocs/stream/pipeline.ex
```diff
@@ -0,0 +1,111 @@
...
+    producer_name = List.first(Broadway.producer_names(:ocs_pipeline))
+    Kernel.send(producer_name, {:resume_position_update, sequence_number})
+    message
+  end
+
+  # TODO
+  # defp update_ocs_sequence_monitor(records) do
+  #   SequenceMonitor.update_ocs_records(records)
+  #   records
+  # end
```
mathcolo, line 61: question: are we going without a sequence monitor after all?

lib/ocs/stream/pipeline.ex
```diff
@@ -0,0 +1,111 @@
...
+    event_records_count =
+      records
+      |> Enum.flat_map(&parse_records/1)
+      # |> update_ocs_sequence_monitor
+      |> Enum.map(&ocs_message/1)
+      |> Enum.map(&apply(ocs_mod, ocs_func, [&1, now] ++ ocs_args))
+      |> Enum.count()
+
+    log_handled_events(sequence_number, ms_behind, event_records_count)
+    producer_name = List.first(Broadway.producer_names(:ocs_pipeline))
```
mathcolo, line 52: question: `producer_names` isn't guaranteed to return a module—it could be an atom. is there a way we can ensure correctness here?

config/test.exs
```diff
@@ -5,7 +5,14 @@ config :orbit,
   environment: "test",
   release: "test",
   force_https?: false,
-  poll_rtr?: false
+  poll_rtr?: false,
+
+  # OCS Stream
+  handle_ocs_message: {OCS.MessageHandler, :receive, []}
```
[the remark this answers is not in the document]
andrewdolce: Actually this is a mistake on my part. It _should_ be overwritten here. (Also note that for some specific test cases, the test will set a different handler using `Application.put_env`) Nice catch.

docker-compose.yml
andrewdolce: I would say this is not a nit. I did not proliferate the typo to the infra work, so this would have actually caused a problem 🤦

lib/ocs/stream/pipeline.ex
andrewdolce: Not sure I follow. I think the intention is that the returned value is the same as the locally registered process name for the producer? Not sure how to guarantee that though.

config/test.exs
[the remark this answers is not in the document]
andrewdolce: Removed.

lib/ocs/stream/pipeline.ex
andrewdolce: @mathcolo After digging more through the Broadway source, I'm pretty confident that this is correct usage? 

Basically when the Broadway.Topology module starts the producer process within the supervision tree, it generates a process name to be registered as part of the producer's child spec. That same name is then stored in configuration (by default, using `:persistent_term`) to capture the topology. The `Broadway.producer_names` function retrieves those stored names, so it should match up.

I've not found a way to, say, override what the producer's name ends up being from the outside, so the best we can do is rely on Broadway to give us the correct name? (This is also what RTR does, and I think how BroadwayKinesis is intended to be used.)

But let me know if you have thoughts/concerns, or if I'm misunderstanding your question.

mathcolo (approved):
Yeah, I'm comfortable with this!

PR #338, opened 2025-06-20
feat: create sidebar

cho-jos:
Asana Task: [Create sidebar](https://app.asana.com/1/15492006741476/project/1200273269966439/task/1210172293290202?focus=true)

Added `<SideBar />` component.
added a new `<LadderPage />` wrapper component that will be in charge of rendering `<Ladders />` and `<SideBar />`.

OUTSTANDING ITEMS:
- the ticket specifies putting this feature behind a keycloak group. currently I have Orbit checking for the group `orbit-rl-sidebar`, but if the group ends up being named something different, then `js/groups.ts` must be updated to reflect that. `fake_oidcc` is also updated with this group as an option for local testing. If the user has the group, then they can click on train pills to make the sidebar "slide-in". If they do not have the user group, then clicking on the train pills does nothing (they still look clickable though because they're `<button />`'s... is there a better way to do this?)
- ^ related to above, need to coordinate with infra/product on Red Line staff that should be included in this initial test group.
- there are still some browser window sizes that throws off the sidebar "pushing" the ladders to the right in order to make space for itself. It works for the most part in a lot of cases, but in the screenrecording below, ~1:05 I try to demonstrate that depending on the size of the browser window (and possibly also the scroll position?) the sidebar will simply slide-out on top of the ladder, instead of "pushing" the ladder and making space. I had a hard time playing around in Tailwind trying to get this to work as robustly as I wanted so this probably could use a small rework. Take a look at 

Checklist

- Tests:
  - `(X)` Has tests
  - `( )` Doesn't need tests
- Product/Design sign off:
  - `(X)` No review needed

### DIFF
--- a/js/components/ladderPage/ladder.tsx
+++ b/js/components/ladderPage/ladder.tsx
@@ -7,6 +7,7 @@ import {
   vehiclesFromPositionsAndTripUpdates,
 } from "../../models/vehicle";
 import { StopStatus } from "../../models/vehiclePosition";
+import { SideBarSelection } from "../sideBar/sidebar";
 import { height } from "./height";
 import { Train } from "./train";
 import {
@@ -14,9 +15,15 @@ import {
   TrainThemes,
   trainThemesByRoutePattern,
 } from "./trainTheme";
-import { ReactElement } from "react";
+import { Dispatch, ReactElement, SetStateAction } from "react";
 
-export const Ladders = ({ routeId }: { routeId: RouteId }): ReactElement => {
+export const Ladders = ({
+  routeId,
+  setSideBarSelection,
+}: {
+  routeId: RouteId;
+  setSideBarSelection: Dispatch<SetStateAction<SideBarSelection | null>>;
+}): ReactElement => {
   const tripUpdates = useTripUpdates();
   const vehiclePositions = useVehiclePositions();
   const stationLists = Stations[routeId];
@@ -50,7 +57,7 @@ export const Ladders = ({ routeId }: { routeId: RouteId }): ReactElement => {
 
   return (
     <>
-      <span>
+      {/* <span>
         Status:{" "}
         {vehiclePositions !== null && tripUpdates !== null ?
           <>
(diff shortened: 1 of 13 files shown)


js/components/ladderPage/ladderPage.tsx
```diff
@@ -0,0 +1,31 @@
...
+export const LadderPage = ({ routeId }: { routeId: RouteId }): ReactElement => {
+  const [sideBarSelection, setSideBarSelection] =
+    useState<SideBarSelection | null>(null);
+
+  return (
+    <div className="relative flex h-screen items-center justify-center overflow-y-auto">
+      <div
+        className={className([
+          "flex overflow-auto mt-72 transition-all duration-300 ease-in-out",
+          sideBarSelection && "ml-80 2xl:ml-0",
```
cho-jos, line 16: This is where, depending on the size of the window, the sidebar sliding in will cause the ladders to move to the right (via. `ml-80` in order to make space for the sidebar. This doesn't work as robustly as I want, especially when you play around with window size and scroll positions and I wanted to specifically call this out as needing some more attention

cho-jos:
not sure why there are front end tests failing in CI, everything works locally for me. Seems like they have to do with test env being unable to import `sideBar/` stuff

js/components/ladderPage/ladder.tsx
```diff
@@ -49,43 +56,29 @@ export const Ladders = ({ routeId }: { routeId: RouteId }): ReactElement => {
...
-                key={index}
-                ladderConfig={stationList}
-                vehicles={vehicles}
-              />
-            ),
-          )}
-        </div>
-      </div>
-    </>
+    <div className="relative flex flex-grow min-[1200px]:justify-center overflow-x-auto snap-x snap-mandatory">
```
andrewdolce, line 59: Not sure I understand why, but removing `justify-center` seemingly helps with the sidebar overlap problem?

mathcolo: It's thresholded, and needed because we center-align at big sizes and left-align on mobile 👍 check out HEAD, it's okay now I think

PR #349, opened 2025-07-01
feat: match VehiclePositions, TripUpdates, and OCS trips on the server

mathcolo:
Asana Task: Part of [Sidebar: Current trip section (data from OCS only)](https://app.asana.com/1/15492006741476/project/1200273269966439/task/1210172293290205)

In order to bring OCS data to the frontend, we could either:
1. Send OCS data over websocket to the client, where the client could then figure out which OCS trip belongs to a particular pill, [or]
2. Change what we're sending over websocket to link OCS trips and VehiclePositions intrinsically.

This is [2]. So this PR:

- Changes the frontend to, instead of accepting independent `VehiclePosition` and `TripUpdate`, accept a `Vehicle` instead
- Adds a `TripMatcherServer` that connects `VehiclePosition`, `TripUpdate`, and `OCS.Trip` together for delivery to the frontend
- Makes it so the `TripMatcherServer` emits new data to the frontend whenever we get new `VehiclePositions`, and if not, pushes new data every 5 seconds.

Checklist

- Tests:
  - `(x)` Has tests
  - `( )` Doesn't need tests
- Product/Design sign off:
  - `( )` No review needed


lib/orbit/vehicle.ex
```diff
@@ -0,0 +1,18 @@
+defmodule Orbit.Vehicle do
+  alias Orbit.Ocs.Trip
+  alias Realtime.Data.TripUpdate
+  alias Realtime.Data.VehiclePosition
+
+  @type t :: %__MODULE__{
+          position: VehiclePosition.t(),
+          trip_update: TripUpdate.t(),
```
andrewdolce, line 8: **question:** Should we declare nullability here with an explicit `| nil`?

lib/realtime/trip_matcher.ex
```diff
@@ -0,0 +1,42 @@
...
+        end)
+
+      ocs_current_and_next =
+        Enum.reduce(ocs_trips, %{current: nil, next: []}, fn trip, acc ->
+          if trip.train_uid == vehicle_id do
+            # credo:disable-for-next-line
+            case acc.current do
+              nil -> %{current: trip, next: acc.next}
+              _ -> %{current: acc.current, next: [trip | acc.next]}
+            end
```
andrewdolce, line 29: **question:** Does this assume that `ocs_trips` is ordered chronologically so that the first found trip is the "current"? Can we guarantee that they aren't out of order, and/or contain past trips?

(This doesn't need to be addressed now. Realizing this PR doesn't yet hook up the actual subscription to the ocs trips table, so maybe questions of ordering/filtering trips will be solved via the DB querying.)

lib/realtime/trip_matcher.ex
```diff
@@ -0,0 +1,42 @@
...
+  alias Orbit.Ocs.Trip
+  alias Orbit.Vehicle
+  alias Realtime.Data.TripUpdate
+  alias Realtime.Data.VehiclePosition
+
+  @spec match_trips([VehiclePosition.t()], [TripUpdate.t()], [Trip.t()]) :: [Vehicle.t()]
+  def match_trips(vehicle_positions, trip_updates, ocs_trips) do
+    Enum.map(vehicle_positions, fn vp ->
+      vehicle_id =
+        case vp.vehicle_id do
```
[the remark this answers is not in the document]
mathcolo: Your assumption is good. We don't need to consider nonrev vehicles though, as nonrev vehicles still have a `x-` prefix on their ids! 😃

lib/orbit/vehicle.ex
mathcolo: Yes! thanks 65f7ea8

test/realtime/trip_matcher_test.exs
```diff
@@ -0,0 +1,119 @@
...
+                 [
+                   %Trip{
+                     train_uid: "5483E6C7",
+                     uid: "1234FFAC"
+                   }
+                 ]
+               )
+    end
+
+    test "finds other OCS trips by vehicle_id (train_uid)" do
```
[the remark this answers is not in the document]
mathcolo: Ah, this is due to pattern matching— it's a `=` instead of a `==`— so while the `VehiclePosition` is there, I'm just not checking for it. 78134cb should make it more clear 😃

andrewdolce (approved):

PR #350, opened 2025-07-03
feat: sidebar est arrival

cho-jos:
Asana Task: [Sidebar: estimated arrival](https://app.asana.com/1/15492006741476/project/1200273269966439/task/1210390452284240?focus=true)

Uses the predicted arrival in the `StopTimeUpdate` sub struct (part of `TripUpdate`). Had to add a new version of `TripUpdates_enhanced.json` for tests since the existing one's structure was a little stale.

**NOTE:** currently there are no `passthrough_time`'s available for terminal stops on nonrev trips: discussed here https://mbta.slack.com/archives/C03K6NLKKD1/p1751485987558089

Checklist

- Tests:
  - `(X)` Has tests
  - `( )` Doesn't need tests
  - `( )` Tests deferred (with justification)
- Product/Design sign off:
  - `( )` Okayed the plan for the feature (e.g. the design files, or the Asana task)
  - `( )` Reviewed the feature as implemented (e.g. on dev-green, or saw screenshots)
  - `(X)` No review needed

PR #352, opened 2025-07-07
feat: current trip section with data from socket (NOT estimated arrival)

mathcolo:
Asana Task: [Sidebar: Current trip section (data from OCS only)](https://app.asana.com/1/15492006741476/project/1200273269966439/task/1210172293290205)

image

Estimated arrival is in https://github.com/mbta/orbit/pull/350

Checklist

- Tests:
  - `(x)` Has tests
  - `( )` Doesn't need tests
- Product/Design sign off:
  - `( )` No review needed


js/components/ladderPage/sidebar.tsx
```diff
@@ -56,15 +58,24 @@ const CurrentTrip = ({ selection }: { selection: SideBarSelection }) => {
       <div className="flex justify-between mt-3">
         <div className="flex flex-col justify-between">
           <span className="text-gray-300">Departure</span>
-          <span>---</span>
+          <span>{getNameForId(current?.originStation) ?? "---"}</span>
           <span className="text-gray-300 mt-5">Arrival</span>
-          <span>---</span>
+          <span>{getNameForId(current?.destinationStation) ?? "---"}</span>
```
andrewdolce, line 63: **question**: Not sure if this scenario is correct/makes sense, but In the event that the `current` trip is missing, but `tripUpdate` is somehow present, could we get fallback values for these from the `tripUpdate.stopTimeUpdates` and/or route pattern?

andrewdolce (approved):
A few nits/questions, but nothing blocking.

js/components/ladderPage/sidebar.tsx
mathcolo: We could, but is that predicated on RTR knowing something we don't?

PR #354, opened 2025-07-09
feat(ocs): persist and recover kinesis stream resume position

andrewdolce:
Asana Task: [OCS Data: Persist and recover Kinesis resume position](https://app.asana.com/1/15492006741476/project/1200273269966439/task/1210733014239778?focus=true)

This relies on a pending change in broadway_kinesis: https://github.com/mbta/broadway_kinesis/pull/3 .

With this change, the `Orbit.Ocs.Stream.Pipeline` module now maintains and persists the kinesis stream "resume position" corresponding to the most recently processed batch of records. This should provide two benefits:

1. The `BroadwayKinesis.Producer` is now notified of the resume position after each batch of messages is processed, and will now resume from that position after reconnecting to the shard (which happens every 5 mins, according to the expectations of `SubscribeToShard`.) Previously it would just use the `:latest` option, which resumes from the most recent message in the shard, potentially introducing gaps.
2. The resume position is also recovered from the DB when the pipeline starts, meaning that we should avoid missing messages during periods of application downtime.

About missing/stale resume positions:

- If there is no stored resume position (e.g. fresh start in new environment, or stream name has changed), we instead resume using a timestamp of midnight for the current day, (ie a few hours before expected start of the current service date.)
  -  This is a divergence from RTR, which would instead use `:latest`, ie potentially missing a bunch of relevant messages for the current date.
- Similarly, if the stored resume position is very stale (based on timestamp of last message when it was stored), then instead we resume from midnight of the current service date.
  - This is a divergence from RTR, which would just use the old sequence number. Kinesis would replay old messages from that sequence (or from the trim horizon, ie oldest stored message from 24 hours ago, if the sequence is older than that.)
  - This means in some cases we are intentionally skipping messages from the prior service date, as they are no longer relevant. However if we prefer that Orbit always try to recover from trim horizon in this case and process the old messages (so as to have a more accurate historical view in the DB), we can talk about that.

Other implementation notes:

- I've implemented this via a Postgres table, which is a departure from RTR's approach (where they use a generic `PersistentState` macro to allow GenServers to automatically persist their state to S3). Should we eventually implement a sequence counter for the OCS-specific sequence numbers (note: different than the Kinesis stream position), I am thinking we could potentially put that information in the same table.
- I have _not_ refactored out our reliance on asking `Broadway` to give us the producer's process name, which @mathcolo I know was a concern of yours previously. I have yet to find a better way to do that (although I did find one alternative that I feel is equally bad).

Checklist

- Tests:
  - `(X)` Has tests
  - `( )` Doesn't need tests
- Product/Design sign off:
  - `( )` No review needed


lib/orbit/ocs/stream/pipeline.ex
```diff
@@ -96,4 +114,87 @@ defmodule Orbit.Ocs.Stream.Pipeline do
...
+    expired = :gt == DateTime.compare(start_of_service_date(), timestamp)
+
+    if expired do
+      Logger.warning("Last OCS Kinesis timestamp #{timestamp} is older than cutoff #{cutoff}.")
+    end
+
+    expired
+  end
+
+  defp start_of_service_date do
```
mathcolo, line 164: suggestion: this name slightly suggests the cutoff time on a particular service date, perhaps `midnight_on_service_date` would be better?

lib/orbit/ocs/stream/pipeline.ex
```diff
@@ -96,4 +114,87 @@ defmodule Orbit.Ocs.Stream.Pipeline do
...
+  # Kinesis Stream Persistence
+
+  defp get_producer_opts(opts) do
+    resume_position = load_resume_position()
+    opts ++ [state: %{resume_position: resume_position}]
+  end
+
+  @spec load_resume_position() :: BroadwayKinesis.SubscribeToShard.starting_position()
+  defp load_resume_position do
+    cutoff = start_of_service_date()
```
mathcolo, line 127: suggestion: could this instead be `midnight`?

cho-jos, line 127: I dig the approach of using the `resume_position` and falling back to the `cutoff` if it's not available

lib/orbit/ocs/stream/pipeline.ex
```diff
@@ -96,4 +114,87 @@ defmodule Orbit.Ocs.Stream.Pipeline do
...
+
+    with stream_name when not is_nil(stream_name) <- get_stream_name(),
+         stream_state when not is_nil(stream_state) <- load_stream_state(stream_name),
+         false <- expired?(stream_state, cutoff) do
+      Logger.info("Will resume OCS Kinesis stream from #{stream_state.resume_position}")
```
andrewdolce, line 132: Realizing I should splunkify these logs better. I'll push a change for that.

lib/orbit/ocs/stream/pipeline.ex
```diff
@@ -96,4 +114,87 @@ defmodule Orbit.Ocs.Stream.Pipeline do
...
+    end
+
+    expired
+  end
+
+  defp start_of_service_date do
+    Util.Time.current_service_date()
+    |> DateTime.new!(
+      ~T[00:00:00.000],
+      Util.Time.current_timezone()
```
mathcolo, line 168: question: sorry, why again is `DateTime.new!` getting `@timezone` and then getting shifted?

lib/orbit/ocs/stream/pipeline.ex
andrewdolce: Renamed in 6c4e4e9

lib/orbit/ocs/stream/pipeline.ex
andrewdolce: Fixed in 6c4e4e9

lib/orbit/ocs/stream/pipeline.ex
```diff
@@ -41,14 +44,21 @@ defmodule Orbit.Ocs.Stream.Pipeline do
       records
       |> Enum.flat_map(&parse_records/1)
       # |> update_ocs_sequence_monitor
-      |> Enum.map(&ocs_message/1)
+      |> Enum.flat_map(&ocs_message/1)
```
mathcolo, line 47: question: would you mind providing the backstory on this?

andrewdolce: As per our conversation, this was an (arguably confusing) way to filter out non-ocs messages. I put up a revised version in 663f02f that I think is more clear.

lib/orbit/ocs/stream/pipeline.ex
```diff
@@ -39,16 +42,24 @@ defmodule Orbit.Ocs.Stream.Pipeline do
 
     event_records =
       records
-      |> Enum.flat_map(&parse_records/1)
+      |> Stream.flat_map(&parse_records/1)
       # |> update_ocs_sequence_monitor
-      |> Enum.map(&ocs_message/1)
+      |> Stream.filter(&ocs_message?/1)
+      |> Enum.map(&unwrap_ocs_data/1)
```
mathcolo, line 48: suggestion: probably favor `Enum.filter` here since the `Enum.map` that immediately follows is not lazy?

lib/orbit/ocs/stream/pipeline.ex
```diff
@@ -96,4 +119,87 @@ defmodule Orbit.Ocs.Stream.Pipeline do
...
+
+    expired
+  end
+
+  @spec persist_resume_position(String.t(), DateTime.t()) :: :ok
+  defp persist_resume_position(resume_position, last_message_timestamp) do
+    stream_name = get_stream_name()
+
+    if stream_name != nil do
+      {:ok, _} =
```
mathcolo, line 183: suggestion: this fails fast, but do we want to bubble up a `:error` to `handle_message`?

lib/orbit/ocs/stream/pipeline.ex
```diff
@@ -76,8 +95,20 @@ defmodule Orbit.Ocs.Stream.Pipeline do
     |> List.wrap()
   end
 
-  @spec ocs_message(map) :: String.t()
-  defp ocs_message(%{"type" => "com.mbta.ocs.raw_message", "data" => %{"raw" => message}}),
+  @spec ocs_message?(map) :: boolean()
+  defp ocs_message?(message)
```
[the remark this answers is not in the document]
andrewdolce: Strictly speaking, we don't, but I was under the impression that it's considered stylistically preferable? I suppose because it indicates to someone looking at the header/spec to expect multiple function patterns, as opposed to just one? That said, I don't feel super strongly about that, so let me know if you'd prefer I remove it / avoid them.

PR #355, opened 2025-07-09
feat: map 15xx trains to 25xx

cho-jos:
Asana Task: [Show 15xx RL trains as 25xx](https://app.asana.com/1/15492006741476/project/1200273269966439/task/1210454158108078?focus=true)

Screenshot 2025-07-09 at 5 17 19 PM

Checklist

- Tests:
  - `(X)` Has tests
  - `( )` Doesn't need tests
- Product/Design sign off:
  - `(X)` No review needed

### DIFF
--- a/js/models/vehiclePosition.ts
+++ b/js/models/vehiclePosition.ts
@@ -22,6 +22,19 @@ export const stopStatusFromData = (data: StopStatusData): StopStatus => {
   }
 };
 
+// remap 15xx trains to 25xx
+export const remapCarLabels = (cars: CarId[]) => {
+  const newCars = [];
+  for (const car of cars) {
+    if (car.startsWith("15")) {
+      newCars.push("2" + car.slice(1));
+    } else {
+      newCars.push(car);
+    }
+  }
+  return newCars;
+};
+
 export const VehiclePositionData = z.object({
   route_id: z.string(),
   direction: z.number(),
@@ -60,8 +73,8 @@ export const vehiclePositionFromData = (
 ): VehiclePosition => ({
   routeId: data.route_id,
   directionId: data.direction,
-  label: data.label,
-  cars: data.cars,
+  label: data.label.startsWith("15") ? "2" + data.label.slice(1) : data.label,
+  cars: remapCarLabels(data.cars),
   stationId: data.station_id,
   stopId: data.stop_id,
   stopStatus: stopStatusFromData(data.current_status),
--- a/js/test/models/vehiclePosition.test.ts
+++ b/js/test/models/vehiclePosition.test.ts
@@ -76,4 +76,36 @@ describe("vehiclePositionFromData", () => {
       tripId: "68077971",
     });
   });
+
+  test("parses 15xx trains as 25xx from raw data", () => {
+    expect(
+      vehiclePositionFromData({
+        route_id: "Red",
+        direction: 0,
+        label: "1566",
+        cars: ["1566", "1567", "1877", "1876", "1839", "1838"],
+        position: { latitude: 42.37469, longitude: -71.11877 },
+        heading: 350,
+        station_id: "place-portr",
+        stop_id: "70065",
+        current_status: "INCOMING_AT",
+        timestamp: "2025-04-29T20:39:49Z",
+        vehicle_id: "R-5482AC4E",
+        trip_id: "68077971",
+      }),
+    ).toEqual({
+      routeId: "Red",
+      directionId: Direction.Southbound,
+      label: "2566",
+      cars: ["2566", "2567", "1877", "1876", "1839", "1838"],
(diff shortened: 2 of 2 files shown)


js/test/models/vehiclePosition.test.ts
```diff
@@ -76,4 +76,36 @@ describe("vehiclePositionFromData", () => {
       tripId: "68077971",
     });
   });
+
+  test("parses 15xx trains as 25xx from raw data", () => {
```
cho-jos, line 80: wasn't sure if this warranted a whole test dedicated to it vs. just integrating some "15xx" trains in the above tests and verifying the output is "25xx" trains. just LMK either way

js/models/vehiclePosition.ts
```diff
@@ -22,6 +22,19 @@ export const stopStatusFromData = (data: StopStatusData): StopStatus => {
   }
 };
 
+// remap 15xx trains to 25xx
+export const remapCarLabels = (cars: CarId[]) => {
+  const newCars = [];
```
cho-jos, line 27: tried to think of a way to do this in-place but ended up just creating a new array since we don't have that many real-world vehicles. open to any and all amendments

js/models/vehiclePosition.ts
```diff
@@ -22,6 +22,19 @@ export const stopStatusFromData = (data: StopStatusData): StopStatus => {
...
+  for (const car of cars) {
+    if (car.startsWith("15")) {
+      newCars.push("2" + car.slice(1));
+    } else {
+      newCars.push(car);
+    }
+  }
+  return newCars;
```
mathcolo, line 35: This might be simpler as:
```
cars.map(c => c.replace(/^15/, "25"));
```

js/test/models/vehiclePosition.test.ts
```diff
@@ -76,4 +76,36 @@ describe("vehiclePositionFromData", () => {
...
+    expect(
+      vehiclePositionFromData({
+        route_id: "Red",
+        direction: 0,
+        label: "1566",
+        cars: ["1566", "1567", "1877", "1876", "1839", "1838"],
```
mathcolo, line 86: Should we do this on the backend instead? What are the tradeoffs?

js/models/vehiclePosition.ts
```diff
@@ -60,8 +73,8 @@ export const vehiclePositionFromData = (
 ): VehiclePosition => ({
   routeId: data.route_id,
   directionId: data.direction,
-  label: data.label,
-  cars: data.cars,
+  label: data.label.startsWith("15") ? "2" + data.label.slice(1) : data.label,
+  cars: remapCarLabels(data.cars),
```
mathcolo, line 77: Should we do this on the backend instead? What are the tradeoffs?

js/models/vehiclePosition.ts
```diff
@@ -60,8 +73,8 @@ export const vehiclePositionFromData = (
 ): VehiclePosition => ({
   routeId: data.route_id,
```
mathcolo, line 74: We probably want to check the route?

js/models/vehiclePosition.ts
cho-jos: My initial thinking was that this was more of a "superficial" change where Orbit knows trains are actually 15xx, but will display them to users as 25xx, so wanted to keep this remapping as just client-side trickery -- but I guess it doesn't matter whether Orbit perceives them as 25xx in the backend or not, probably actually better to keep it all aligned since that is ultimately how they will exist in Orbit's domain

js/models/vehiclePosition.ts
cho-jos: wow, huge oversight by me. thanks for the callout

js/models/vehiclePosition.ts
cho-jos: decided to handle this in Elixir instead, utilizing the built in `String.replace_prefix`, lmk any thoughts

js/components/ladderPage/train.tsx
```diff
@@ -26,6 +26,11 @@ export const Train = ({
   setSideBarSelection: Dispatch<SetStateAction<SideBarSelection | null>>;
 }): ReactElement => {
   const orientation = forceDirection == 0 ? "right-0" : "left-0";
+  const label = vehicle.vehiclePosition.label;
+  const displayLabel =
+    vehicle.vehiclePosition.routeId === "Red" && label.startsWith("15") ?
+      "2" + label.slice(1)
```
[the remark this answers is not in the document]
cho-jos: hmm perhaps I was trying too hard with the whole "15xx right up to the point that we need to display in the frontend". I was trying to avoid having something like a processing function in the `vehiclePosition` front-end model but maybe that's what should happen here since it _is_ a front-end model

PR #364, opened 2025-07-15
feat: sidebar next trip section

cho-jos:
Asana Task: [Sidebar: Next trip section](https://app.asana.com/1/15492006741476/project/1200273269966439/task/1210172293290208?focus=true)

2 thoughts to note:
- Initially I thought about combinging "Current" and "Next" trips into one "Trip" component to render, but I liked how keeping them separate makes the code faster/easier to read. I think this also lends itself to better organization in the future as we add more features/enhancements to one specific area without needing to touch the other -- however I am open to any thoughts about consolidating them into one component that will contain both the Current and Next Trips' details.
- I reorganized the sidebar tests + integrated next trip testing into the tests that were already written to minimize code reuse (otherwise, it was the same exact tests copy pasted for 2 similar, but different components).

Screenshot 2025-07-15 at 5 13 02 PM

Checklist

- Tests:
  - `(X)` Has tests
  - `( )` Doesn't need tests
- Product/Design sign off:
  - `(X)` No review needed


js/test/components/ladderPage/sidebar.test.tsx
```diff
@@ -19,17 +19,18 @@ describe("sidebar", () => {
...
+      <SideBar
+        selection={{ vehicle: vehicleFactory.build() }}
+        close={() => {}}
+      />,
+    );
+    expect(view.getByText("Current Trip")).toBeInTheDocument();
+    expect(view.getByText("Next Trip")).toBeInTheDocument();
+  });
 
+  describe("Trips", () => {
```
andrewdolce, line 33: **suggestion**: I think this new "Trips" section makes sense for testing for commonalities between the two components, but I notice that the "Offset" tests still live only under "Current Trip section". Might make sense to move that here and add test coverage for next trip offsets, given that the components are separate.

andrewdolce (approved):
Nice! One suggestion, non-blocking.

Also: I could go either way on whether to combine the components or not - There is obviously a lot of commonality, but the spacing and extra fields are different enough (and the components are small enough) that I'm cool with your decision to keep them separate.

js/test/components/ladderPage/sidebar.test.tsx
cho-jos: I don't thing the "next" trips have offsets, according to the side bar [figma](https://www.figma.com/design/BJwHVkEm191yYqbMzk1dT5/Orbit-MVP?node-id=901-21354&t=5EvL8TreIcHP83jW-0), perhaps I'm misunderstanding/not seeing something there?

PR #368, opened 2025-07-16
feat(ocs): revise message timestamp handling

andrewdolce:
Continued work on Asana Task: [Read in the OCS raw messages](https://app.asana.com/1/15492006741476/project/1210239137170761/task/1209929021017070?focus=true)

For context: OCS messages contain an `hh:mm:ss` timestamp that indicates when the relevant change occurred in OCS. Since these timestamps are not fully specified, in that they don't tell you the actual _date_, the parser must infer that.

Two changes here:

- The parser would previously infer the relevant date based on _current time_. If we are receiving messages in realtime, this is reasonable, but not if Orbit is processing old messages from Kinesis. It makes more sense to use the timestamp that the CloudEvent was emitted by Trike, since that's a more accurate indicator of when OCS actually emitted this change.

- I've taken a stab at revising how the timestamps are parsed and reasoned about. Some details:
  - RTR considers any time >= 1 hour in the future to be stale and assumes instead that the time corresponds to yesterday. I've kept this logic in place, albeit now relative to event emission time, not current time at parsing. (Keep in mind that currently we don't expect to see messages older than 24 hours, since that's how long Kinesis keeps them. Maybe it's possible for OCS to _emit_ very old messages, but hopefully not.)
  - There are some (maybe) nitpicky scenarios where I think the original parsing code introduces small errors during daylight savings time. This mostly would occur for stale message processing, and maybe isn't super important, but I've tried to fix it anyway.
  - I've added tests, particularly around D.S.T. change boundaries.
  - I've removed our use of Timex, at least in this function, as I believe Elixir's recent updates to the native `DateTime` now provide what we need.

Still TODO in future work:

- I have not yet changed how *schedule times* within a message are interpreted relative to the service date. I may revise that or at least add more tests.
- We may want to combine some of our time utils into a single module (although maybe not when you consider that we might extract the OCS-specific stuff into to a library or lift it upstream into Trike.)

Checklist

- Tests:
  - `(X)` Has tests
  - `( )` Doesn't need tests
- Product/Design sign off:
  - `( )` No review needed

### DIFF
--- a/lib/orbit/ocs/utilities/time.ex
+++ b/lib/orbit/ocs/utilities/time.ex
@@ -5,6 +5,89 @@ defmodule Orbit.Ocs.Utilities.Time do
   require FastLocalDatetime
   require Logger
 
+  @doc """
+  Parse the internal timestamp of an OCS message, and interpret it into a full DateTime,
+  based on the DateTime when the OCS message was emitted.
+
+  msg_time: An (HH:mm:ss)-formatted string used internally by OCS to indicate when the
+            message change occurred.
+  emitted_datetime: The DateTime that TID systems received the event from OCS. Trike
+                    currently emits these in UTC.
+  timezone: The preferred timezone for the output.
+  """
+  @spec interpret_ocs_message_timestamp(String.t(), DateTime.t(), String.t()) :: DateTime.t()
+  def interpret_ocs_message_timestamp(
+        msg_time,
+        emitted_datetime,
+        timezone \\ Application.get_env(:orbit, :timezone)
+      ) do
+    # Interpret message time as local time for the same date that the message was emitted.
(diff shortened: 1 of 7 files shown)


lib/orbit/ocs/utilities/time.ex
```diff
@@ -5,6 +5,89 @@ defmodule Orbit.Ocs.Utilities.Time do
...
+  defp closer_datetime(earlier, later, datetime) do
+    diff = DateTime.diff(later, earlier, :second)
+    cutoff = DateTime.shift(earlier, second: trunc(diff / 2))
+
+    if DateTime.before?(datetime, cutoff) do
+      earlier
+    else
+      later
+    end
+  end
```
mathcolo, line 89: question: would this be simpler as
```
Enum.min_by([earlier, later], &DateTime.diff(datetime, &1))
```
?

mathcolo (approved):

PR #370, opened 2025-07-17
feat: add VehicleEventDetector

mathcolo:
Asana Task: [Sidebar: actual departure (track departures automatically)](https://app.asana.com/1/15492006741476/project/1200273269966439/task/1209646113903412)

This PR:
- Brings a heavy rail version of VehicleEventDetector over to Orbit
- ⚠️ Does not yet record or facilitate recording `actual_departure` anywhere. That's tomorrow!

Checklist

- Tests:
  - `(x)` Has tests
  - `( )` Doesn't need tests
- Product/Design sign off:
  - `( )` No review needed

### DIFF
--- a/lib/orbit/rail_line.ex
+++ b/lib/orbit/rail_line.ex
@@ -31,4 +31,7 @@ defmodule Orbit.RailLine do
       _ -> :none
     end
   end
+
+  @spec from_route_id(Realtime.Data.route_id()) :: t()
+  def from_route_id(:Red), do: :red
 end
--- a/lib/realtime/data/stations.ex
+++ b/lib/realtime/data/stations.ex
@@ -56,4 +56,58 @@ defmodule Realtime.Data.Stations do
       "Braintree-02" => "place-brntn"
     }
   end
+
+  @spec next_stations(String.t() | nil, integer()) :: [String.t()]
+  def next_stations(station_id, direction) do
+    case {station_id, direction} do
+      # Southbound
+
+      {"place-alfcl", 0} -> ["place-davis"]
+      {"place-davis", 0} -> ["place-portr"]
+      {"place-portr", 0} -> ["place-harsq"]
+      {"place-harsq", 0} -> ["place-cntsq"]
+      {"place-cntsq", 0} -> ["place-knncl"]
+      {"place-knncl", 0} -> ["place-chmnl"]
+      {"place-chmnl", 0} -> ["place-pktrm"]
+      {"place-pktrm", 0} -> ["place-dwnxg"]
+      {"place-dwnxg", 0} -> ["place-sstat"]
+      {"place-sstat", 0} -> ["place-brdwy"]
+      {"place-brdwy", 0} -> ["place-andrw"]
+      {"place-andrw", 0} -> ["place-jfk"]
+      {"place-jfk", 0} -> ["place-shmnl", "place-nqncy"]
+      {"place-shmnl", 0} -> ["place-fldcr"]
+      {"place-fldcr", 0} -> ["place-smmnl"]
+      {"place-smmnl", 0} -> ["place-asmnl"]
+      {"place-asmnl", 0} -> []
+      {"place-nqncy", 0} -> ["place-wlsta"]
+      {"place-wlsta", 0} -> ["place-qnctr"]
+      {"place-qnctr", 0} -> ["place-qamnl"]
+      {"place-qamnl", 0} -> ["place-brntn"]
+      {"place-brntn", 0} -> []
+      # Northbound
+      {"place-alfcl", 1} -> []
+      {"place-davis", 1} -> ["place-alfcl"]
+      {"place-portr", 1} -> ["place-davis"]
+      {"place-harsq", 1} -> ["place-portr"]
+      {"place-cntsq", 1} -> ["place-harsq"]
+      {"place-knncl", 1} -> ["place-cntsq"]
+      {"place-chmnl", 1} -> ["place-knncl"]
+      {"place-pktrm", 1} -> ["place-chmnl"]
+      {"place-dwnxg", 1} -> ["place-pktrm"]
+      {"place-sstat", 1} -> ["place-dwnxg"]
+      {"place-brdwy", 1} -> ["place-sstat"]
+      {"place-andrw", 1} -> ["place-brdwy"]
+      {"place-jfk", 1} -> ["place-andrw"]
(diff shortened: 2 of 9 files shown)


lib/realtime/data/stations.ex
```diff
@@ -56,4 +56,58 @@ defmodule Realtime.Data.Stations do
       "Braintree-02" => "place-brntn"
     }
   end
+
+  @spec next_stations(String.t() | nil, integer()) :: [String.t()]
+  def next_stations(station_id, direction) do
+    case {station_id, direction} do
+      # Southbound
+
+      {"place-alfcl", 0} -> ["place-davis"]
```
mathcolo, line 65: I had no simple ideas for computing this at runtime, so I did what Glides does and put it together manually. Makes me wonder, though...

lib/orbit/rail_line.ex
```diff
@@ -31,4 +31,7 @@ defmodule Orbit.RailLine do
       _ -> :none
     end
   end
+
+  @spec from_route_id(Realtime.Data.route_id()) :: t()
```
mathcolo, line 35: At a later date we may want to remove `Realtime.Data.route_id()`, or more clearly delineate it from `RailLine`.

lib/realtime/vehicle_event_detector.ex
```diff
@@ -0,0 +1,380 @@
...
+    # list of {old_vp | nil, new_vp | nil} pairs
+    Util.List.zip_by(old_vehicle_positions.entities, new_vehicle_positions.entities, fn vp ->
+      MapSet.new(vp.cars)
+    end)
+    |> Enum.flat_map(&vehicle_events_for_one_train(&1, service_date))
+  end
+
+  @spec vehicle_events_for_one_train({VehiclePosition.t(), VehiclePosition.t()}, Date.t()) :: [
+          StationEvent.t()
+        ]
```
andrewdolce, line 143: **suggestion**: Maybe the type spec here should represent that the tuples can contain `nil`?

lib/realtime/vehicle_event_detector.ex
```diff
@@ -0,0 +1,380 @@
...
+    Enum.map(station_events, fn station_event ->
+      %VehicleEvent{
+        service_date: service_date,
+        cars: new_vp.cars,
+        station_id: station_event.station_id,
+        vehicle_id: new_vp.vehicle_id,
+        rail_line: RailLine.from_route_id(new_vp.route_id),
+        direction_id: station_event.direction,
+        arrival_departure: station_event.arrival_departure,
+        timestamp: new_vp.timestamp
```
andrewdolce, line 156: **question**: Does this timestamp need to be sanitized to comply with Ector's `utc_datetime`? Or is there some upstream reason that we can rely on it to already be in an acceptable format?

lib/realtime/vehicle_event_detector.ex
```diff
@@ -0,0 +1,380 @@
...
+    Enum.flat_map(@terminals, fn {pullout, departure_direction} ->
+      if departure_direction != nil && new_vp.direction == departure_direction &&
+           new_vp.station_id in Stations.next_stations(pullout, departure_direction) do
+        case new_vp.current_status do
+          :STOPPED_AT ->
+            # The train wasn't registered until it made it to the next station.
+            [
+              StationEvent.new(pullout, departure_direction, :departure),
+              StationEvent.new(new_vp.station_id, departure_direction, :arrival)
+            ]
```
andrewdolce, line 196: **question**: Does this mean that the original departure event will be timestamped with the same time as the arrival event for the station where this vehicle appeared? I suppose that's maybe the best we can do?

priv/repo/migrations/20250717170938_vehicle_events.exs
```diff
@@ -0,0 +1,24 @@
...
+      add(:service_date, :date, null: false)
+      add(:cars, :jsonb, null: false)
+      add(:station_id, :string, null: false)
+      add(:vehicle_id, :string, null: false)
+      add(:direction_id, :integer, null: false)
+      add(:rail_line, :rail_line, null: false)
+      add(:arrival_departure, :arrival_departure, null: false)
+      add(:timestamp, :utc_datetime, null: false)
+
+      timestamps()
```
andrewdolce, line 19: **question**: So I have not yet fixed our other entities to use non-naive datetimes in their auto-gen'd timestamps, as we had at one point discussed. I don't know if it makes sense to do that now for this particular entity, or just change them all later should we decide that's still needed.

test/realtime/vehicle_event_detector_test.exs
```diff
@@ -0,0 +1,770 @@
...
+            current_status: :IN_TRANSIT_TO,
+            timestamp: 80
+          })
+        ]
+      }
+
+      new_vehicle_positions = %{
+        timestamp: 200,
+        entities: [
+          # Train didn't move, should not make a new departure event at Riverside
```
andrewdolce, line 109: **nit**: Shawmut, not Riverside

lib/realtime/data/stations.ex
andrewdolce: Haven't internalized GTFS enough to know if it easily represents these station connections? In any case, I think it's fine to hard-code it like this.

andrewdolce (commented):
Nice! I like the way the different movement cases are broken out, and feels like we can be pretty confident in all of that since it's already working in Glides.

Few comments, nothing major. (My question about `VehicleEvent.timestamp` being safe for Ecto is maybe the one important thing.)

priv/repo/migrations/20250717170938_vehicle_events.exs
mathcolo: As we'll probably need to end up doing them at all together, I figure we should keep it consistent for now and do a big one-off alter + configuration adjustment? (But soon, while tables are small...)

lib/realtime/vehicle_event_detector.ex
mathcolo: Indeed this works as is because [rtr.ex's use of `from_unix!()`](https://github.com/mbta/orbit/blob/4748d809182ae9e5023b87bdb4973d8c5f7b8ec6/lib/realtime/rtr.ex#L41) always [gives a UTC stamp](https://hexdocs.pm/elixir/1.18.4/DateTime.html#from_unix/3).

andrewdolce (approved):

PR #377, opened 2025-07-25
feat: track sidebar events in Fullstory

andrewdolce:
Asana Task: [Tracking: fullstory event on sidebar open](https://app.asana.com/1/15492006741476/project/1200273269966439/task/1210572330010013?focus=true)

Description TODO

Checklist

- Tests:
  - `(X)` Has tests
  - `( )` Doesn't need tests
- Product/Design sign off:
  - `( )` No review needed


js/telemetry/trackingEvents.ts
```diff
@@ -0,0 +1,55 @@
...
+import { OCSTrip } from "../models/ocs";
+import { estimatedArrivalOfVehicle } from "../models/vehicle";
+import { FullStory } from "@fullstory/browser";
+
+export enum FullStoryEventName {
+  SideBarOpened = "Preston test event",
+  // TODO: Change to
+  // SideBarOpened = "Orbit: Ladder Side Bar Opened"
+}
+
```
andrewdolce, line 11: I will change this to the real event name before merge. Please let me know if anyone has feedback on the actual name to be used. I am considering prefixing it with "Orbit:" as shown, since events in Fullstory are shared across the whole org and there doesn't seem to be a good way to organize them by application in certain contexts.

js/models/vehicle.ts
```diff
@@ -50,3 +50,10 @@ export const vehicleFromVehicleData = (vehicleData: VehicleData): Vehicle => {
     },
   };
 };
+
+export const estimatedArrivalOfVehicle = (vehicle: Vehicle) => {
```
mathcolo, line 54: observation: ha, I did this too. https://github.com/mbta/orbit/pull/378/files#diff-38050556722f95612a65c8d9f25ffda52df105a5c1e14a989b83724d8559acb0R67

Not sure how much the array bounds check matters?

js/telemetry/trackingEvents.ts
mathcolo: Seems good to me 👍 `[Application]:` is also what we do in Splunk for the same reason.

js/telemetry/trackingEvents.ts
```diff
@@ -0,0 +1,55 @@
...
+export const trackSideBarOpened = (selection: SideBarSelection) => {
+  const { vehicle } = selection;
+  const currentTrip = vehicle.ocsTrips.current;
+  const nextTrip = vehicle.ocsTrips.next[0] as OCSTrip | undefined;
```
mathcolo, line 15: question: Why was this needed?

js/models/vehicle.ts
andrewdolce: The main difference is that yours returns explicit `null`, whereas mine is less specific and can return `undefined`. So I think I prefer yours. Since that branch is about to go in, I'll wait and rebase.

js/telemetry/trackingEvents.ts
mathcolo: nice! yeah, I like this better 👍

PR #379, opened 2025-07-28
fix: use utc for inserted_at/updated_at in all schemas

andrewdolce:
Asana Task: 🛠 [Investigate Ecto timestamp timezones](https://app.asana.com/1/15492006741476/project/1200882337457260/task/1210835734470501?focus=true)

I know the ticket just said to investigate, but I ended up writing a migration as part of testing this out.

I can post a more detailed writeup, but in short:
- Ecto defaults to `:naive_datetime` for `timestamps()`, which stores no timezone info and is cast to `NaiveDatetime`. However, the values written to the table do happen to match expected time in UTC.
- This PR updates schemas to use `:utc_datetime`, which is identical in the table, but more explicit w.r.t. intentions, and makes Ecto cast to fully specified `DateTime`.
- For correctness, I included an alter for all existing tables. It is somewhat nominal, since there is no actual column type change, and does not need to convert values, since the values are already "UTC".

Checklist

- Tests:
  - `( )` Has tests
  - `(X)` Doesn't need tests
- Product/Design sign off:
  - `( )` No review needed

### DIFF
--- a/lib/orbit/authentication/user.ex
+++ b/lib/orbit/authentication/user.ex
@@ -20,7 +20,7 @@ defmodule Orbit.Authentication.User do
     field(:email, :string)
     field(:permissions, {:array, UserPermission})
     field(:groups, {:array, :string}, virtual: true)
-    timestamps()
+    timestamps(type: :utc_datetime)
   end
 
   def changeset(struct, params \\ %{}) do
--- a/lib/orbit/badge_serial.ex
+++ b/lib/orbit/badge_serial.ex
@@ -13,7 +13,7 @@ defmodule Orbit.BadgeSerial do
     belongs_to(:employee, Employee)
     field(:badge_serial, :string)
 
-    timestamps()
+    timestamps(type: :utc_datetime)
   end
 
   def changeset(badge_serial, attrs \\ %{}) do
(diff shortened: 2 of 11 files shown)


lib/orbit/authentication/user.ex
```diff
@@ -20,7 +20,7 @@ defmodule Orbit.Authentication.User do
     field(:email, :string)
     field(:permissions, {:array, UserPermission})
     field(:groups, {:array, :string}, virtual: true)
-    timestamps()
+    timestamps(type: :utc_datetime)
```
andrewdolce, line 23: An alternative way to configure this is to use the `@timestamps_opts ` [schema attribute](https://hexdocs.pm/ecto/Ecto.Schema.html#module-schema-attributes), which is meant to be called after `use Ecto.Schema` but before the call to `schema`.

The only advantage I can see to doing it that way is we could actually write our own wrapper macro, something like `use Orbit.Ecto.Schema`, which combines the two so we wouldn't need to repeat this for every call to `timestamps`. But I didn't go that route because I think maybe that obscures what's happening unless someone knows to go look inside that macro.

mathcolo (approved):

PR #382, opened 2025-07-30
feat: show OCS "last updated" timestamp at bottom of sidebar

andrewdolce:
Asana Task: [Sidebar: last updated statement](https://app.asana.com/1/15492006741476/project/1200273269966439/task/1210164685019845?focus=true)

Shows the last update time from OCS data (based on current or next trip `updated_at`, whichever is newer.)

A few other semi-related changes that I made to get the layout as desired:
- I tweaked the height of the header (and Orbit logo icon), since I noticed it did not match current [Figma](https://www.figma.com/design/BJwHVkEm191yYqbMzk1dT5/Orbit-MVP?node-id=752-5474&t=5lXc4xHiFrII8atT-0)
- I used the header height to explicitly size/position the side bar, since before the header was actually overlapping the side bar slightly. I did this without really altering the current approach to the header (which is to have it overlap page content with z-index). Maybe we can discuss the intentions there.
- I changed side panel to use flex column layout, and wrapped most of the content in another div (which takes up maximum possible height) in order to allow the timestamp to be pushed to the bottom.
  - To make this slightly more readable, I factored the consist UI into another component.

(I suppose technically it would be cleaner for me to break out some of these sidebar / header changes into a separate pre-req PR, and can do that if preferred. The commits are already mostly split out, but not sure if we care.)

Checklist

- Tests:
  - `(X)` Has tests
  - `( )` Doesn't need tests
- Product/Design sign off:
  - `( )` No review needed


js/models/vehicle.ts
```diff
@@ -132,3 +132,14 @@ export const lateForNext = (vehicle: Vehicle): number | null => {
 
   return minutesAfter(estimated, nextDeparture);
 };
+
+export const latestOcsUpdatedAt = (vehicle: Vehicle): DateTime | null => {
+  const datetimes = [vehicle.ocsTrips.current, vehicle.ocsTrips.next[0]]
```
mathcolo, line 137: question: necessary to check the length of `next`?

js/test/components/ladderPage/sidebar.test.tsx
```diff
@@ -103,6 +103,22 @@ describe("sidebar", () => {
...
+        const view = render(
+          <SideBar
+            selection={{
+              vehicle: vehicleFactory.build(),
+            }}
+            close={() => {}}
+          />,
+        );
+        const lastUpdated = view.getByText(
+          /Last updated from OCS trainsheets at 2:00a/,
```
mathcolo, line 117: suggestion: is this the full text of the element? it might not need to be an expression.

js/components/ladderPage/sidebar.tsx
```diff
@@ -23,37 +24,46 @@ export const SideBar = ({
...
+        <NextTrip vehicle={selection.vehicle} />
       </div>
-      <CurrentTrip vehicle={selection.vehicle} />
-      <NextTrip vehicle={selection.vehicle} />
+      <LastOcsUpdated vehicle={selection.vehicle} />
     </aside>
   );
 };
 
+const Consist = ({ vehicle }: { vehicle: Vehicle }) => {
```
mathcolo, line 45: question: was this just a code style change, or is there a visual treatment change associated with the consist?

js/models/vehicle.ts
andrewdolce: I personally think it's OK as-is, but I can add an explicit check if you prefer? If `next[0]` is `undefined`, it gets stripped out in the filter call. (Also note that there is a unit test case that exercises that situation.)

js/components/ladderPage/sidebar.tsx
andrewdolce: Code style. Basically because I ended up nesting most of the side bar content inside another `div`, I thought it read better if I split this out to keep it shorter.

js/models/vehicle.ts
mathcolo: oh! 🤦 it's fine as is, then

PR #389, opened 2025-08-01
fix: avoid event tracking crash if FS is not initialized

andrewdolce:
This is a follow-up for Asana Task: [Tracking: fullstory event on sidebar open](https://app.asana.com/1/15492006741476/project/1200273269966439/task/1210572330010013?focus=true)

Custom event tracking for the sidebar relies on Orbit being configured with the FS org ID, which can be set locally in `dev.secret.exs`. If this ID is missing, FS is not initialized, which is fine, except that this causes the side bar event to crash, preventing the side bar from opening.

This adds an explicit check for initialization, and if not, skips the event. 

Checklist

- Tests:
  - `(X)` Has tests
  - `( )` Doesn't need tests
- Product/Design sign off:
  - `(X)` No review needed

### DIFF
--- a/js/telemetry/trackingEvents.ts
+++ b/js/telemetry/trackingEvents.ts
@@ -1,12 +1,19 @@
 import { SideBarSelection } from "../components/ladderPage/sidebar";
 import { estimatedArrival } from "../models/tripUpdate";
-import { FullStory } from "@fullstory/browser";
+import {
+  FullStory,
+  isInitialized as isFullStoryInitialized,
+} from "@fullstory/browser";
 
 export enum FullStoryEventName {
   SideBarOpened = "Orbit: Ladder Side Bar Opened",
 }
 
 export const trackSideBarOpened = (selection: SideBarSelection) => {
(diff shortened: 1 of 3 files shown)


js/telemetry/trackingEvents.ts
```diff
@@ -1,12 +1,19 @@
...
 
 export enum FullStoryEventName {
   SideBarOpened = "Orbit: Ladder Side Bar Opened",
 }
 
 export const trackSideBarOpened = (selection: SideBarSelection) => {
+  if (!isFullStoryInitialized()) {
+    return;
+  }
+
```
mathcolo, line 16: nit: would it be better for us to wrap `FullStory()` so further tracking functions in this file don't need to repeat this?

andrewdolce: @mathcolo I think I follow? Took a stab at that in 5d72dda. See what you think. My only criticism with it is that now the individual event functions may be doing extra unnecessary work leading up to the FS call, rather than doing the init check first as an early-out.

PR #502, opened 2026-06-09
fix: Estimated Arrival Time during diversion

rycollins:
Asana Task: [🐞 Orbit 'Estimated' arrival not working during RL diversion](https://app.asana.com/1/15492006741476/project/1200882337457260/task/1214494378895416?focus=true)

Updated estimated arrival time to show the time for the STU that matches the destination station from OCS. If there isn't a match or OCS doesn't return destination data, show "---" as before.

Checklist
- Appearance:
  - `( )` Light & dark mode
- Browsers:
  - `(x)` Chromium
  - `( )` Firefox
  - `( )` Safari
- Privacy:
  - `( )` PR description free of internal data
  - `( )` Logging free of internal data
- Tests:
  - `(x)` Has tests
  - `( )` Doesn't need tests
- Product/Design sign off:
  - `( )` No review needed

### DIFF
--- a/js/components/ladderPage/sidebar.tsx
+++ b/js/components/ladderPage/sidebar.tsx
@@ -1,10 +1,8 @@
-import { formatStationName, gtfsToOcsStationName } from "../../data/stations";
+import { DateTime } from "luxon";
+import { formatStationName, gtfsToOcsStationName, ocsStationNameToGtfs } from "../../data/stations";
 import { dateTimeFormat } from "../../dateTime";
 import { CarId } from "../../models/common";
-import {
-  estimatedArrival,
-  lastArrivalStationId,
-} from "../../models/tripUpdate";
+import { estimatedArrival } from "../../models/tripUpdate";
 import {
   lateArrival,
   lateDeparture,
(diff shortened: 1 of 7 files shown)


js/models/tripUpdate.ts
```diff
@@ -64,19 +66,14 @@ export const TripUpdateData = z.object({
...
 export type TripUpdateData = z.infer<typeof TripUpdateData>;
 
-export const estimatedArrival = (tu?: TripUpdate): DateTime | null => {
-  if (tu == undefined || tu.stopTimeUpdates.length === 0) {
-    return null;
-  }
+export const estimatedArrival = (vehicle: Vehicle): DateTime | null => {
+  const gtfsStopId = ocsStationNameToGtfs(
+    vehicle.ocsTrips.current?.destinationStation ?? "",
+  );
```
andrewdolce, line 72: **question**: One of the pre-existing behaviors here was that if for some reason we had no trip information from OCS, but did have a `tripUpdate` via RTR, that we'd just default to showing the estimated arrival time of the final stop in RTR. With this change, we'd instead show nothing, correct?

I guess my question is also, "did we talk about this?" I think I missed when this ticket was refined... I honestly could see an argument for "just show nothing" in that scenario, but would maybe want to clear that with product/design.

js/data/stations.ts
```diff
@@ -237,9 +237,11 @@ export const formatStationName = (station: string | null | undefined) => {
...
 
-export const gtfsToOcsStationName = (stationId: string) => {
-  return Object.values(Stations)
-    .flat()
-    .flat()
-    .find((station) => station.id === stationId)?.ocs_station_name;
+export const ocsStationNameToGtfs = (ocsStationName: string): string | null => {
+  return (
+    Object.values(Stations)
+      .flat(2)
```
andrewdolce, line 243: **praise**: Huh. Somehow never noticed there's a `depth` param on `flat()`. Neat!

js/components/ladderPage/sidebar.tsx
```diff
@@ -77,29 +75,7 @@ const Consist = ({ vehicle }: { vehicle: Vehicle }) => {
 const CurrentTrip = ({ vehicle }: { vehicle: Vehicle }) => {
   const current = vehicle.ocsTrips.current;
 
-  // The last prediction in the TripUpdate (i.e. the destination) must match the OCS
-  // trip. Before this, vehicles at Ashmont SB that are assigned to the next NB
-  // trip in OCS, but not RTR yet (due to the turnaround), will have an old estimated arrival
-  // https://app.asana.com/1/15492006741476/project/1206105669438487/task/1210824268353890
```
andrewdolce, line 83: **comment**: Not sure that you need to change anything here, because I think it's OK, but...

As per this comment, we at one point decided to check that the trip from OCS and the trip from RTR had matching destinations, because of scenarios where a vehicle was turning around, in which OCS might have moved to the next trip before RTR, leading to a mismatch.

The new method searches the list to find a matching station, which is great. But in theory two mismatched trips could contain the same station, say somewhere in the middle of the line?

Practically speaking though, RTR drops station estimates as the trip progresses, ie only shows the remaining stops. So a vehicle turning around would not likely have an RTR `tripUpdate` containing the destination of the next trip. So again, I think this change is probably OK. Mostly calling this out as something we should keep an eye on. If we start seeing evidence of mismatches, we may need some new criteria to check.

andrewdolce (commented):
Nice! Searching the list of stops feels a lot less fragile.

Left one comment about handling the case where the OCS trip is missing. We may actually not want to do that anymore, but wanted to double check.

js/models/tripUpdate.ts
rycollins: Yep, with this change an estimated arrival will never show if we don't have trip info from OCS.

Can't remember for sure if this was stated explicitly, good callout. Will double check with product/design.

My interpretation comes from the Asana task calling out not to show the Estimated arrival time for a station when it doesn't match what the Arrival column says, and the arrival station name in the sidebar currently refers to the current OCS trip, no check against the values in stopUpdates.

js/data/stations.ts
rycollins: Also thought this was neat! Stumbled on it when I was double checking what `flat` does on MDN

js/components/ladderPage/sidebar.tsx
rycollins: Good point, didn't give full consideration to the implications of this (RTR dropping station estimates as the trip progresses wasn't something I was actually aware of, though I can see that now looking at the example in the Asana task). 

Even if it's probably OK as is, as a thought experiment to see if I'm understanding this case properly, could we check OCS departure/arrival stations against the direction in tripUpdate to confirm? Or is there a simpler way to validate that OCS and RTR are actually referring to the same trip?

andrewdolce: @rycollins In terms of "how to tell if the OCS trip and RTR trip are the same"... unfortunately I think the most straightforward approach, something like "just compare the IDs" doesn't work? Because OCS refers to trips via UIDs from its system, whereas RTR generates and maps those to public-facing IDs for GTFS. And I don't believe there's an easy way to go between them without knowledge of RTR's internal state. Maybe worth double checking with transit data team to make sure I'm not wrong / haven't overlooked something obvious? 

But regardless, this is starting to feel like out-of-scope w.r.t. your ticket? Wondering if it would be better to open a tech debt ticket to capture this.

cho-jos: just chiming in to say that @andrewdolce you are right and the OCS trip uid <-> gtfs trip id connection is not straightforward and the logic that rtr executes to make the connection is complex. I asked about it while doing the "orange line in glides" spike relevant slack thread [here](https://mbta.slack.com/archives/C06D663HRRN/p1779902827820689).

andrewdolce (approved):
Based on Dave M's comment about the missing OCS data case, it sounds like this is good to go as-is.

js/models/tripUpdate.ts
rycollins: Discussed with Dave/Zoe in comments on Asana task. Verdict is we can leave this blank and may revisit attempting to use RTR data only if OCS is absent (see [comment](https://app.asana.com/1/15492006741476/task/1214494378895416/comment/1215805856888262?focus=true))

PR #516, opened 2026-06-25
feat(ladder): add ladder configurations for OL and BL

mathcolo:
Asana Task: [🛰️ 📐 Set up static data for OL/BL](https://app.asana.com/1/15492006741476/project/1200882337457260/task/1214913596677802)

Prepare to expand to OL and BL by adding ladder configuration for both lines. Unused.

~⚠️ Note/TODO: before merging, I need to verify which platform at Oak Grove and Forest Hills is which (left/right).~
done

Checklist

- Appearance **N/A:**
  - `( )` Light & dark mode
- Browsers **N/A:**
  - `( )` Chromium
  - `( )` Firefox
  - `( )` Safari
- Privacy:
  - `(x)` PR description free of internal data
  - `(x)` Logging free of internal data
- Tests:
  - `( )` Has tests
  - `(x)` Doesn't need tests
- Product/Design sign off:
  - `( )` No review needed


js/data/stations.ts
```diff
@@ -208,6 +208,280 @@ export const Stations: Record<RouteId, LadderConfig[]> = {
...
+        id: "place-ogmnl",
+        name: "Oak Grove",
+        ocs_station_name: "OAK GROVE",
+        spacingRatio: 2.5,
+        location: { latitude: 42.437735, longitude: -71.070875 },
+        stop_ids: ["70036", "Oak Grove-01", "Oak Grove-02"],
+        forcedDirections: new Map([
+          // TODO: verify platforms
+          ["Oak Grove-01", 1],
+          ["Oak Grove-02", 0],
```
andrewdolce: **question**: Is it that you need to verify which platform ID is which w.r.t. direction?

**question**: Realizing that I don't actually understand the need for encoding a forced direction of the child stop IDs, vs just using vehicle direction like we do for most general stops? Are vehicle directions known to be untrustworthy at terminals?

andrewdolce (approved):
One question, but overall LGTM.

js/data/stations.ts
mathcolo: Exactly, this is because the direction is explicitly _not_ useful at terminals; STOPPED_ATs get the departing direction. But that breaks emulated-real-life ladder rendering, since one of the platforms is actually on the other side. So we force one of the platforms to look like it's in the wrong direction.

PR #524, opened 2026-06-30
chore(deps): update dependencies to address vulnerabilities

andrewdolce:
Asana Task: None

This addresses vulnerabilities flagged by hex. Many of the problems are with libraries included as sub-dependencies, so I needed to update the top-level dependencies pulling them in.

A list of what changed:

| Vulnerability | Updated | Fix |
| ------------  | -------- | --- |
|`bandit` 1.10.1 | 1.12.0 | update at top level |
|`cowlib` 1.0.2  | 2.18.0 | revert `laboratory`, update `cowboy` to 2.17.0 |
| `decimal` 2.3.0 | 3.1.1 | update `ecto_sql` --> 3.14.0, `jason` --> 1.4.5, drop `geocalc` |
| `mint` 1.7.1 | 1.9.0 | update pinned version |
| `plug` 1.19.1 | 1.20.2 | update pinned version  |
| `postgrex` 0.22.0 | 0.22.0| update at top level |

Note that to update `cowlib` I had to update `laboratory`, which is a bit of a special case, since we are including it as a git dependency. The history there:
- We were at one point using a branch in Paul Swartz's fork of the library, which specified `cowboy` as a development-only dependency. But it turns out that the prod build actually needs it, so if it's not already being included via other deps (as in Glides) then this causes the build to fail.
- To fix that, I created my own fork to promote `cowboy` to be a prod dependency. But that fixes it at ~> 1.0.
- Updating cowboy inside laboratory is proving troublesome because of other dependency conflicts. So the fix here was to revert to an older version of Paul's fork wherein cowboy is a dev-only dependency, and install an updated version in our top-level dependencies, essentially as a "co-dependency". I will be looking into ways to fix this longer term. 

Checklist

- Appearance: (N/A)
- Browsers:
  - `(X)` Chromium
  - `( )` Firefox
  - `( )` Safari
- Privacy:
  - `(X)` Commits free of internal data
  - `(X)` PR description free of internal data
  - `(X)` Logging free of internal data
- Tests:
  - `( )` Has tests
  - `(X)` Doesn't need tests
- Product/Design sign off:
  - `(X)` No review needed


cho-jos (approved):
LGTM

PR #541, opened 2026-07-23
feat: add car search

mathcolo:
Asana Task: [🪐 📐 Orbit car search function](https://app.asana.com/1/15492006741476/project/1200882337457260/task/1215731340693515)

image

Checklist

- Appearance:
  - `( )` Light & dark mode
- Browsers:
  - `( )` Chromium
  - `( )` Firefox
  - `( )` Safari
- Privacy:
  - `(x)` PR description free of internal data
  - `(x)` Logging free of internal data
- Tests:
  - `(x)` Has tests
  - `( )` Doesn't need tests
- Product/Design sign off:
  - `( )` No review needed


js/components/ladderPageShared/ladderPage.tsx
```diff
@@ -43,6 +48,73 @@ export const LadderPage = ({ routeId }: { routeId: RouteId }): ReactElement => {
...
+          !newPill &&
+          selection !== null &&
+          selection.searchedCar === undefined &&
+          sideBarSelection?.searchedCar !== undefined
+        ) ?
+          {
+            ...selection,
+            searchedCar: sideBarSelection.searchedCar,
+          }
+        : selection;
```
andrewdolce: **question**: If I'm following, this covers the specific case of having opened the sidebar via car search, and then explicitly clicking on the same pill that is already opened. So the sidebar is essentially "reopened", but we don't want to lose the car highlight. Is that correct?

I would wonder if this could be solved more simply by just making a repeat selection of the same car a no-op?

js/components/ladderPageShared/ladderPage.tsx
```diff
@@ -43,6 +52,73 @@ export const LadderPage = ({ routeId }: { routeId: RouteId }): ReactElement => {
...
+        return;
+      }
+
+      const match = findVehicleMatch(vehicles, query.trim());
+      if (match === null) {
+        setSideBarSelection(null);
+      }
+    },
+    [setSearchQuery, sideBarSelection, vehicles],
+  );
```
andrewdolce, line 79: **question**: The extra call to `findVehicleMatch` in here is so that if the sidebar is already in the state of having selected a car via search, interacting with the query at all with immediately trigger a new search, correct? I think that's fine but wondering if that's necessary. Is this a design AC or does it just feel wrong if we don't auto-trigger the search in this state?

andrewdolce (approved):
LGTM! Couple of questions, but nothing blocking.

js/components/ladderPageShared/ladderPage.tsx
mathcolo: That's a fantastic idea, changed! 31147e2 thanks

js/components/ladderPageShared/ladderPage.tsx
mathcolo: This is actually here because otherwise changing "1952" to "952" with a single backspace would trigger sidebar close/new search, which seems weird since it's still a valid car. Generally speaking there aren't single keystrokes one would make that cause a valid query to remain valid, so I'll ask design to see if we can ignore the edge case.

mathcolo: Per [comment](https://app.asana.com/1/15492006741476/task/1215731340693515/comment/1216965180329026?focus=true), any modification in the search bar counts as clearing the result. Changed.

